PR #1, opened 2017-03-01
a little patch for apache on sles12, a big patch for syslinux-6.04-pre

crbaird:
Hi -
   I wanted to test the waters to see how you guys feel about including syslinux-6.04-pre with warewulf-3.7. I'm sure you're aware of the bug preventing PXE+UEFI provisioning with syslinux-6.03, and I'm happy to report that in our testing, 6.04-pre works out of the box for both legacy BIOS and UEFI booting in our CI infrastructure. The only gotcha is that for some reason UEFI is taking about a minute and a half longer to boot a node than legacy. Any thoughts? 

Thanks


gmkurtzer:
Is this a pre-release for syslinux or considered stable? Is that what OpenHPC is using?

crbaird:
It is technically pre-release. According to the changelog from March of 2016:

At the time of writing this temporary list, version 6.04 is not yet released; there are no specific plans, no road-map, no particular TODO list, no ETA.

Barring any unforeseen unpleasantness, we intend to ship it in OpenHPC 1.3.

gmkurtzer:
I've been bitten by weird edge cases of some syslinux versions not working with some firmware and thus having to provide, multiple versions, so I'm a bit cautious.

Have you guys already been testing it for OHPC?

crbaird:
Understood. It's been in our broadwell CI for a couple of days now in legacy mode, and we're currently adding an EFI job there. Testing it on other hardware would certainly behoove us.

gmkurtzer:
I would prefer to wait until it is either formally released or some decent testing has been done. With that said, lets approach it from the other perspective. Is there a tangible need to upgrade (e.g. is the released version failing in some way or do we need new features like EFI support)?

Thanks!

crbaird:
That's perfectly reasonable. I am unaware of any problems with the released version aside from the broken EFI PXE support, which at least a few folks are interested in.

gmkurtzer:
How urgent is the EFI support needed as you see it? Right now I think most BIOSs offer a legacy boot mode, but will that always exist?

Have you tested EFI boot through syslinux? The last time I tried it was broken bad (with the current version we have released). lol

crbaird:
We've tested EFI booting, and it works just fine out of the box with the caveat that single node boots take about a minute longer.

koomie:
Hey @gmkurtzer. Just to expand on the question on EFI support, we have received  questions/requests/ and some problems around the lack of EFI support to date since UEFI is showing up as the default config on some newer systems. Also, legacy mode is indeed, no longer a guarantee. As an example, KNL nodes from certain vendors do not have legacy mode. Similarly, the ARM nodes that were donated for the community CI environment also don't have legacy mode. 

So, my .02 is that I think it is a necessity at some point to be compatible with newer platforms.  As @crbaird said, we have it enabled now on the CI system for x86 with the syslinux 6.04pre and it's going ok for UEFI and legacy PXE installs.

bensallen:
We require UEFI booting for Intel Adam's Pass KNL systems and HPE's Apollo systems so far. UEFI will be our default method of booting for any new hardware. OmniPath HCAs appear to ship only with a UEFI network driver, and don't include an option ROM for legacy PXE boot.

Has anyone asked the Syslinux folks for an ETA on 6.04? 

We've been using 6.04-pre's EFI binaries since August. However we continue to use 6.0.3's lpxelinux.0 for legacy boot, so I can't comment on if 6.03 is stable or not for legacy boot.

koomie:
I can't comment on 6.04-pre's stability for legacy boot 100% either, but we did start using it in the OpenHPC CI system yesterday, and haven't noticed anything abnormal when using legacy thus far.

gmkurtzer:
Let's go ahead with the syslinux version included in this PR. We can always back it out if we get complaints and do a point release.

@crbaird Do you have any other changes to include on this PR before I merge?

Thanks everyone!

crbaird:
Nope- we're pretty happy with this one. Thanks.

gmkurtzer:
Merged! Thanks!

PR #5, opened 2017-04-17
Upgrade ipmitool to 1.8.18, add openssl-devel BuildRequire

bensallen:
Grabbed the latest ipmitool version. Also lanplus support requires openssl-devel to be on the system to compile, so added a BuildRequire the ipmi spec file.


gmkurtzer:
Looks good to me, thanks @bensallen!

PR #6, opened 2017-04-21
Add new database.conf option 'database chunk size'

bensallen:
Adds a new database.conf option 'database chunk size' to override of database's max_allowed_packet. This is needed in our case as our MySQL db has a very large value set, 1073741824. With this large value any time Warewulf pulls an image it uses a considerable amount of memory on the DB and the Warewulf server.


gmkurtzer:
Looks good to me too! Thanks!

PR #7, opened 2017-05-08
Rev busybox to 1.26.2

bensallen:
Rev busybox to 1.26.2, update and merge busybox config with new defconfig based settings, remove unneeded or unused patches.


bensallen:
@jmstover Originally it looks like we had a patch that put wget's output onto stdout instead of stderr. I believe this would have hid the output but probably logged it? Since that patch doesn't apply cleanly and we haven't been using it for a while anyways, I added -q to the vnfs wget download to hide the progress bar from the console.

jmstover:
Figured that was probably the reason, sounds fine by me... I just like the bar as "sometimes" transfers will hang, and that's a decent visual indicator. Not like I can't modify it to keep it. ;)

PR #8, opened 2017-05-15
Ubuntu support

macdems:
I have added support for Ubuntu Xenial. With the proposed changes I am able to create VNFS and provision it successfully to diskless machines.

This PR also fixes https://github.com/warewulf/warewulf3/issues/2.

### DIFF
--- a/provision/initramfs/init
+++ b/provision/initramfs/init
@@ -351,7 +351,7 @@ if [ -n "$WWPOSTSHELL" -a "$WWPOSTSHELL" != "0" ]; then
     setsid /bin/cttyhack /bin/sh
 fi
 
-if [ -x "$NEWROOT/sbin/init" ]; then
+if [ -x "$NEWROOT/sbin/init" ] || [ -h "$NEWROOT/sbin/init" -a -x "$NEWROOT`readlink $NEWROOT/sbin/init`" ]; then
     msg_white "Stopping syslogd: "
     if killall syslogd >/dev/null 2>&1; then
         wwsuccess
--- a/vnfs/bin/wwvnfs
+++ b/vnfs/bin/wwvnfs
@@ -363,6 +363,7 @@ vnfs_create_file($)
     if (exists($exclude_map{"$path"})) {
         &dprint("Excluding: $path\n");
     } else {
+        $path = encode_utf8($path);
         print $filepipe ".$path\n";
    }
 
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,17 @@ provision/warewulf-provision-*/
 cluster/warewulf-cluster-*/
 ipmi/warewulf-ipmi-*/
 vnfs/warewulf-vnfs-*/
+provision/3rd_party/_work/
+provision/3rd_party/ldlinux.c32
+provision/3rd_party/ldlinux.e32
+provision/3rd_party/ldlinux.e64
+provision/3rd_party/lpxelinux.0
+provision/3rd_party/prepare
+provision/3rd_party/pxelinux.0
+provision/3rd_party/syslinux32.efi
+provision/3rd_party/syslinux64.efi
+provision/initramfs/_work/
+provision/initramfs/capabilities/*/capability.cpio
+provision/initramfs/capabilities/*/rootfs/
+provision/initramfs/initramfs.cpio
+provision/initramfs/rootfs/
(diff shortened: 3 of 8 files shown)


jmstover:
Hi @macdems,

  Thanks for the Ubuntu fixes.

  Was there a reason for removing the `wwrun` function to the ssh-keygen call in 50-ssh_keys.init?

-J

macdems:
Ubuntu Xenial does not support RSA1 keys anymore. So by removing `wwrun` I have made generation of this keys optional: user will be notified with a warning and the keys will not be generated, but the process will continue. The same with ECDSA, which may be not supported on the older systems.

jmstover:
Okay, sounds good. I forgot that wwrun returns with the return value of the process.

@gmkurtzer - I don't see anything off on this. Merge is cool here if you don't see anything jumping out.

provision/initramfs/init
```diff
@@ -351,7 +351,7 @@ if [ -n "$WWPOSTSHELL" -a "$WWPOSTSHELL" != "0" ]; then
     setsid /bin/cttyhack /bin/sh
 fi
 
-if [ -x "$NEWROOT/sbin/init" ]; then
+if [ -x "$NEWROOT/sbin/init" ] || [ -h "$NEWROOT/sbin/init" -a -x "$NEWROOT`readlink $NEWROOT/sbin/init`" ]; then
```
gmkurtzer: We could probably just trust that if `$NEWROOT/sbin/init` is either executable or a symlink we are good. e.g.:

`if [ -x "$NEWROOT/sbin/init" -o -L "$NEWROOT/sbin/init" ]; then`

Thoughts?

vnfs/bin/wwvnfs
```diff
@@ -363,6 +363,7 @@ vnfs_create_file($)
     if (exists($exclude_map{"$path"})) {
         &dprint("Excluding: $path\n");
     } else {
+        $path = encode_utf8($path);
```
gmkurtzer: 👍

gmkurtzer:
Everything looks good to me too, I just have one question in regards to the link test that I commented on in the diff.

Thanks!

provision/initramfs/init
jmstover: Well... the only think I can think of is, by using readlink, you're verifying that the target `$NEWROTO/sbin/init` symlink points to is executable. Speaking of...

@macdems On the list you mentioned changing the symlink to a relative path instead of absolute. I don't see that happening in your `post_chroot()` (include-ubuntu) function. Was that "fix" going to be added in, or was it just mentioned as something to be aware of?

macdems: @gmkurtzer I would rather test if the target is existing and executable.

@jmstover If you do the workaround I mentioned in #2, neither `provision/initramfs/init` nor `provision/initramfs/capabilities/transport-http/wwgetvnfs` need to be chanhged. However I prefer the solution in this pull request, as it does not require changing the chroot. Even if we change the link in the `post_chroot()`, the user may e.g. update the packages in the chroot after some time, and the relative link may be restored back to the absolute one which will prevent booting.

gmkurtzer: I agree that it is smarter to see if the target exists and executable, but to do that right would technically require a recursive loop/test as the link target could also be a link. What about just testing for the existence of `$NEWROOT/sbin/init` and trust that the VNFS knows what it needs internally?

jmstover: Just making sure that `/sbin/init` exists in some form would be fine to me.

jmstover:
I'll go a head an merge this and then change `provision/initramfs/init` to be just a check that `$NEWROOT/sbin/init` exists.

Thanks much @macdems

macdems:
2062c90c68cd8590c8772179b9afe688a80f1f09 makes it buggy again!!! The reason is that

    [ -e "$NEWROOT/sbin/init" ]

follows symlinks. So the test will fail. You need:

    [ -e "$NEWROOT/sbin/init" -o -L "$NEWROOT/sbin/init" ]

I would also suggest changing `provision/initramfs/capabilities/transport-http/wwgetvnfs` for consistency. I have created another pull request with a fix.

gmkurtzer:
I am fine with doing the test like that, but since we are being a bit more specific, I would want it to look like my first suggestion:
```
if [ -x "$NEWROOT/sbin/init" -o -L "$NEWROOT/sbin/init" ]; then
```

macdems:
So this is exactly my latest patch in #9. `-h` and `-L` are equivalent and these are the only tests that do not check for the link target (which will not exist if `/sbin/init` is the absolute link).

PR #9, opened 2017-05-16
Fixed /sbin/init symlink check.

macdems:
2062c90c68cd8590c8772179b9afe688a80f1f09 makes it buggy again!!! The reason is that

    [ -e "$NEWROOT/sbin/init" ]

follows symlinks. So the test will fail. You need:

    [ -e "$NEWROOT/sbin/init" -o -L "$NEWROOT/sbin/init" ]

I would also suggest changing `provision/initramfs/capabilities/transport-http/wwgetvnfs` for consistency. I have created another pull request with a fix.

### DIFF
--- a/provision/initramfs/capabilities/transport-http/wwgetvnfs
+++ b/provision/initramfs/capabilities/transport-http/wwgetvnfs
@@ -33,7 +33,7 @@ while true; do
         WGETEXIT=$?
 
         wait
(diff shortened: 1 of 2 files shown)


jmstover:
I'm dumb... thanks!

macdems:
:+1:

PR #11, opened 2017-05-17
Check $NEWROOT/sbin/init status

jmstover:
Check that `$NEWROOT/sbin/init` is either executable, or a symlink. Not just that it exists.


gmkurtzer:
Looks good to me, thanks!

PR #12, opened 2017-05-24
Integrate Parted and Rewrite 20-filesystems

bensallen:
## Tasks

- [x] Integrate parted into base initrd
- [x] Rewrite 20-filesystems to use parted and parted-like command syntax. Eg:

```
# Pick device that the following commands will run against
select /dev/sda

# Parted specific commands.
mklabel gpt
mkpart primary ext4 0% 100%
mkpart primary linux-swap 20% 50%
mkpart primary ext4 50% 100%
name 1 boot
name 2 swap
name 3 root

# Custom commands
## Formatting
mkfs 1 ext4 boot
mkfs 2 swap swap
mkfs 3 ext4 root

## Create fstab entries and mount
fstab 1 /boot ext4 defaults 0 0
fstab 2 swap swap defaults 0 0
fstab 3 / ext4 defaults 0 0

# Tmpfs root (equivalent to default settings)
select tmpfs
fstab "" / tmpfs rw:relatime:seclabel:mode=555 0 0
```

Example of inputing commands via the object command currently:

```
wwsh -y object modify -a fs='select\ /dev/sda' n01
wwsh -y object modify -a fs="mklabel\ gpt" n01

wwsh -y object modify -a fs="mkpart\ primary\ ext4\ 0%\ 20%" n01
wwsh -y object modify -a fs="mkpart\ primary\ linux-swap\ 20%\ 50%" n01
wwsh -y object modify -a fs="mkpart\ primary\ ext4\ 50%\ 100%" n01

wwsh -y object modify -a fs="name\ 1\ boot" n01
wwsh -y object modify -a fs="name\ 2\ swap" n01
wwsh -y object modify -a fs="name\ 3\ root" n01

wwsh -y object modify -a fs="mkfs\ 1\ ext4\ boot" n01
wwsh -y object modify -a fs="mkfs\ 2\ swap\ swap" n01
wwsh -y object modify -a fs="mkfs\ 3\ ext4\ root" n01

wwsh -y object modify -a fs="fstab\ 3\ /\ ext4\ defaults\ 0\ 0" n01
wwsh -y object modify -a fs="fstab\ 1\ /boot\ ext4\ defaults\ 0\ 0" n01
wwsh -y object modify -a fs="fstab\ 2\ swap\ swap\ defaults\ 0\ 0" n01
```

- [x] Provision CLI updated to import FS command file, eg: ```wwsh provision set [node] --filesystem=cluster.layout```  where 'cluster.layout' is a file
- [x] Provision CLI updated to remove --dformat, --dpartition, and --filesys
- [x] Add better feedback in 20-filesystems for each step, i.e
```
    * Running Parted on <disk>    [OK|FAILED]
    * Formatting <device>         [OK|FAILED]
    * Mounting <mount path>       [OK|FAILED]
```
- [x] Add basic input validation for each command on import of file
- [x] Add example layout files into /etc/warewulf/filesystem/example_gpt.layout, etc
- [x] Parted maintainer sign-off on Warewulf not being a derivative work
- [ ] FSF sign-off on Warewulf not being a derivative work

## Notes

- A select (other than the first instance), mkfs, and fstab command will cause all the previous parted specific commands to be executed in a single call to parted.
- Top-down order of commands matter. 
  - The root partition fstab command should always be specified first.
  - First command must be a select.
- When specifying a virtual filesystem in a fstab command, specify the partition number as an empty string as shown in the tmpfs example above.
- When selecting for a virtual filesystem, use the value that is used in the first column in /etc/fstab, i.e. tmpfs in the tmpfs example above.
- When specifying multiple mount options in the fstab command, use ":" as the delimitation character not a comma.


bensallen:
@gmkurtzer @jmstover, and anyone else this PR is ready to be merged. Please review when you get a chance.

gmkurtzer:
This looks really cool. It seems to me that we are already moving into the provisioning recipe format we were thinking of for WW4.

BTW, I am assuming it does, but how does the object transposition work as it gets rewritten into the node config shell exports? How resilient is the parsing?

bensallen:
@gmkurtzer Yes the recipe format that we have discussed in WW4 and possibly making this work portable to WW4 was in my mind when developing it.

The cmds file is imported line by line. Each line is made into an array entry. nodeconfig.pl joins this array into a comma separated string when fetched. 20-filesystems splits on commas. 

Due to this, fs() in Provision.pm will bail if any commas are included in any commands except fstab. If the fstab options argument has commas they're transformed into colons, and transformed back into commas in 20-filesystems.

Provision.pm also checks that commands are valid from a list have at least enough arguments.

macdems:
`wwsh provision print` shows old filesystems instead of the new format.

macdems:
I would also suggest to change `Provision.pm` so `wwsh provision --filesystem=file` looks for the `file` in `/etc/warewulf/filesystem` if it fails a direct open.

macdems:
@bensallen This modification does not work for me for diskless provisioning and tmpfs.

macdems:
The problem is that in Ubunto bootstap `seclabel` is not supported in initrd. I suggest removing it from the default options.

macdems:
Applying this diff should correct the problem:

    diff --git a/provision/etc/filesystem/tmpfs_example.cmds b/provision/etc/filesystem/tmpfs_example.cmds
    index 21c3bd6..ed3aeb9 100644
    --- a/provision/etc/filesystem/tmpfs_example.cmds
    +++ b/provision/etc/filesystem/tmpfs_example.cmds
    @@ -1,3 +1,3 @@
     # Tmpfs root (equivalent to 20-filesystems' default settings)
     select tmpfs
    -fstab "" / tmpfs rw:relatime:seclabel:mode=555 0 0
    +fstab "" / tmpfs rw:relatime:mode=555 0 0
    diff --git a/provision/initramfs/capabilities/setup-filesystems/20-filesystems b/provision/initramfs/capabilities/setup-filesystems/20-filesystems
    index 2e0a9e2..b47eecd 100644
    --- a/provision/initramfs/capabilities/setup-filesystems/20-filesystems
    +++ b/provision/initramfs/capabilities/setup-filesystems/20-filesystems
    @@ -21,7 +21,7 @@ set -o pipefail
     [ -f /etc/functions ] && . /etc/functions
  
     # By default mount a tmpfs volumne for root
    -WWFS=${WWFS:="select tmpfs,fstab "" / tmpfs rw:relatime:seclabel:mode=555 0 0"}
    +WWFS=${WWFS:="select tmpfs,fstab "" / tmpfs rw:relatime:mode=555 0 0"}
  
     run_parted() {
       local SELECT_DISK="${1}"
    @@ -204,6 +204,9 @@ update_fstab_and_mount() {
           fi
         ;;
  
    +    "")
    +    ;;
    +
         *)
           wwlogger "Unknown parted command: ${FSCMD}"
           exit 2

bensallen:
@macdems Nice catch on seclabel. 

Curious why did you add the case "") in the case ${FSCMD}? Is something adding in empty lines somewhere?

macdems:
@bensallen Theoretically there are no empty lines. But without this I got an error message. I suggest to add this case. it costs nothing and will allow to write more legible commands files (e.g. to separate visually parted commands from fstab).

bensallen:
@macdems Provision.pm which is used by the wwsh provision CLI strips newlines and comments during import of fs cmds. Even via the object command I don't think one can include an empty string in the FS array. Can you grab the output of `wwsh object print -p FS <node>` of the node that has the issue without this extra case?

macdems:
@bensallen Forget it. It is ok now. The problem is that if there is an error, you got two ERROR messages: one from `wwfailure` and one after an empty line, probably generate by the init script. I suggest to add `msg_white " * filesystems"` before each `exit` command even if it is `exit 2`.

macdems:
Just a tiny comment. I find it more convenient to change line 539 of Provision.pm to:

            } elsif (open(FILE, $path) || open(FILE, $fs_cmds_dir . $path . ".cmds")) {

bensallen:
@gmkurtzer and I have agreed that we are good to merge this.

bensallen:
@macdems I agree with that change, just haven't had time to implement it. Likely I'll include the auto suffix of .cmds as you suggest and then add a 3rd condition that lets you use any filename. Thanks.

PR #14, opened 2017-05-25
MySQL and other fixes.

macdems:
This PR fixes an wwinit DATABASE: it will ask for mysql root password if required and it cleans up db uses creation.

The other fix is a unit test fix and and some simple readme to be visible on the main github page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,12 @@
+# Warewulf
+
+Warewulf is a scalable systems management suite originally developed to manage
+large high-performance Linux clusters. Focused on general scalable system
+management, it includes a framework for system configuration, management,
+provisioning/installation, monitoring, event notification, and more via a modular
+plugin architecture. Install the components and features you need or leverage
+the existing system configurations stored within Warewulf to create custom
+solutions to meet your particular needs.
+
+This repository is a fork of https://github.com/warewulf/warewulf3.
+Its development focuses on providing Warewulf support for Ubuntu based systems.
--- a/README.md
+++ b/README.md
@@ -7,6 +7,3 @@ provisioning/installation, monitoring, event notification, and more via a modula
 plugin architecture. Install the components and features you need or leverage
 the existing system configurations stored within Warewulf to create custom
 solutions to meet your particular needs.
-
-This repository is a fork of https://github.com/warewulf/warewulf3.
-Its development focuses on providing Warewulf support for Ubuntu based systems.
--- a/common/libexec/wwinit/10-database.init
+++ b/common/libexec/wwinit/10-database.init
@@ -22,6 +22,7 @@ fi
 wwreqroot
 
 DATASTORE=`egrep "^database driver\s*=" $WAREWULF_SYSCONFDIR/warewulf/database.conf | awk -F = '{print $2}' | sed -e 's/\s*//g'`
+DBSERVER=`egrep "^database server\s*=" $WAREWULF_SYSCONFDIR/warewulf/database.conf | awk -F = '{print $2}' | sed -e 's/\s*//g'`
 DBNAME=`egrep "^database name\s*=" $WAREWULF_SYSCONFDIR/warewulf/database.conf | awk -F = '{print $2}' | sed -e 's/\s*//g'`
 DBUSER=`egrep "^database user\s*=" $WAREWULF_SYSCONFDIR/warewulf/database.conf | awk -F = '{print $2}' | sed -e 's/\s*//g'`
 DBPASS=`egrep "^database password\s*=" $WAREWULF_SYSCONFDIR/warewulf/database.conf | awk -F = '{print $2}' | sed -e 's/\s*//g'`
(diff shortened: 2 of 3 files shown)


common/libexec/wwinit/10-database.init
```diff
@@ -111,24 +122,14 @@ fi
...
-        reply_success
-    else
-        reply_error
-    fi
-fi
-
-if [ -n "$DBROOTUSER" -a "$DBROOTUSER" != "root" ]; then
-    wwprint "Updating database permissions for root user"
-    if echo "GRANT ALL PRIVILEGES ON $DBNAME.* TO $DBROOTUSER@localhost IDENTIFIED BY '$DBROOTPASS';" | mysql $DBNAME; then
+    if echo "GRANT ALL PRIVILEGES ON $DBNAME.* TO $DBUSER IDENTIFIED BY '$DBPASS';" | mysql $DBNAME; then
```
bensallen: Why the change from `$DBROOTUSER@localhost` to `$DBROOTUSER`?

macdems: Because the database may be on a different machine. Actually, the script should be modified to read the host from the configuration file. I will try to correct this.

README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Warewulf is a scalable systems management suite originally developed to manage
+large high-performance Linux clusters. Focused on general scalable system
+management, it includes a framework for system configuration, management,
+provisioning/installation, monitoring, event notification, and more via a modular
+plugin architecture. Install the components and features you need or leverage
+the existing system configurations stored within Warewulf to create custom
+solutions to meet your particular needs.
+
+This repository is a fork of https://github.com/warewulf/warewulf3.
```
bensallen: Can you take off the last paragraph, which appears to be specific to your work? Otherwise this is a good place holder for a README until we get the docs work merged in.

macdems: It has already been changed.

macdems:
I have modified the `10-database.init` script. Now it works if the DB server is on some other machine.

bensallen:
Instead of prompting the user for a password in 10-database.init, I think we should just exit with an error if the value isn't found in database.conf, database-root.conf, or if the credentials fail to work.

macdems:
There problem is `database-root.conf` is incorrectly named, From what I can see it simply contains warewulf user password, which overrides the one in `database.conf`, so it can be hidden with system permissions. It has nothing to do with the root password.

Actually the root password is needed for the `wwinit` only. So I think it makes sense to ask for it if necessary (my version tries to access DB withouth the password first). Otherwise e.g. on Ubuntu one is unable to finish the init process at all.

bensallen:
The point of database-root.conf is to hide a read/write database user via unix permissions and have a read-only user available in database.conf more broadly available. The root in the config file refers to the unix root user not the database. I do admit it's a very coarse permissions model. There's no reason, as you say, for either user to have privileges to create or drop databases or tables, etc.

My issue with prompting is if someone scripts or runs configuration management around wwinit it can hang on the prompt. 

If we do keep the prompt the UI layout should be cleaned up from the examples in #15 that @jtfrey posted, so the text isn't mangled.

macdems:
If run from a script, some `wwinit` config option or an enviromental variable should suffice. However, in normal use `wwinit` is probably run manually just once after installing Warewulf.

common/libexec/wwinit/10-database.init
```diff
@@ -90,45 +115,35 @@ fi
 
 if [ -z "$DBVERS" ]; then
     wwprint "Creating database schema"
-    if echo "$SCHEMA" | mysql "$DBNAME"; then
+    if echo "$SCHEMA" | mysql $DBOPTIONS "$DBNAME"; then
```
gmkurtzer: This should probably be done via a redirect so we don't loose formatting/quoting/escaping through the echo.

macdems: Can you explain? I don't understand what do you mean here?

jtfrey:
Avoid quoting issues by writing the schema to a temp file and then:


if mysql $DBOPTIONS "$DBNAME" < "$SCHEMA_FILE" ; then

jtfrey:
With regard to prompting for a password when creating the database:  there's a general need for some additional options on the `wwinit DATABASE` functionality, I think.

1. The check/install of e.g. "mysql-server" RPMs shouldn't be performed if the database exists on a different server -- rather, you'd want to ensure "mysql" (client) RPMs are installed.  So if "database server" is "localhost" or is omitted from the configuration, one can assume the server RPMs should be checked.  But what if I set "database server" to the public DNS name of the node on which I'm doing `wwinit` -- then it should also be checking for the server RPMs.
2. The "root database user" in the Warewulf sense is not likely the "root user" of the database.  So we have to assume there's a third database user Warewulf _does not know about_, for whom credentials must be provided to do the `wwinit DATABASE` operations.
3. Most databases allow users to set default connection properties and credentials in a file in the user's home directory or in the environment itself.  With the PostgreSQL wwinit functionality in  #14 , the assumption was that the sysadmin running `wwinit` would use that to streamline the passing of credentials.  Only if the first attempt to connect to the (non-local) database server fails do we put up a password prompt.  Personally, I'd rather NOT prompt and require that the person setup the appropriate env vars.
4. CLI flags to override any default behaviors (e.g. --install-db-server-pkgs) might be helpful.

macdems:
@jtfrey I fully agree with all your issue. Actually my patch does try to connect the DB without any prompt first. However, providing the DB administrative user and the password as `wwinit` command line options may be an option (although unsafe, as it allows everybody to see the password easily).

macdems:
Closing this PR as this combines three unrelated issues. I will resubmit it one by one.

common/libexec/wwinit/10-database.init
gmkurtzer: My bad, I forgot we had the entire schema in a variable scalar. It really should be in a file so we can `cat schema.txt | mysql ...`. Using `echo` can be tricky if there are any other quotes, escapes or anything else in the string, as `echo` and/or the shell can parse them.

This is fine considering the simple schema, sorry for the alarm.

PR #15, opened 2017-05-26
Reorganization of Warewulf::DataStore::SQL/* classes

jtfrey:
- New BaseClass.pm defines common functionality among all DBD subclasses
  - *_impl hook functions that subclasses can override without repeating all validation code
- MySQL.pm updated to derive from BaseClass.pm
- MySQL_hybrid.pm : derives from MySQL.pm, overrides binstore to use files in a directory on-disk
- PostgreSQL.pm : derives from BaseClass.pm, implements interface to a PostgreSQL database for warewulf DataStore

Updated Warewulf::DataStore::SQL to do on-the-fly loading of the target SQL/*.pm class (a'la plugins)

The schema for the PostgreSQL database tables is cited at the top of the PostgreSQL.pm file.  No support for initialization of it via wwinit yet (sorry!)


bensallen:
Hi @jtfrey this looks pretty slick. Hopefully will get some time soon to review this and test this out. How much testing have you done around this work?

gmkurtzer:
Indeed, very awesome! I'm looking forward to play with this, probably will happen next week'ish as I'm right up against a Singularity release.

Thanks!

jtfrey:
I've confirmed that all three modules work properly w.r.t.

instantiation and initialization
basic object retrieval, display
adding/updating/removing objects (datastore, lookups)
adding/dumping binstore objects

For the PostgreSQL module I started from a fresh database with the schema newly created in it (see PostgreSQL.pm for that).  For testing the MySQL.pm and MySQL_hybrid.pm I used a database created using the latest OpenHPC-bundled warewulf which was already populated with cluster data.

common/lib/Warewulf/DataStore/SQL/PostgreSQL.pm
```diff
@@ -0,0 +1,294 @@
...
+}
+
+
+=back
+
+=head1 SEE ALSO
+
+Warewulf::ObjectSet Warewulf::DataStore
+
+=head1 COPYRIGHT
```
jmstover: Add yourself in this copyright section as well if you wish since this is the perldoc section.

jmstover:
Looks nice. I'll try testing this some when I get a chance. At least the MySQL parts....

common/lib/Warewulf/DataStore/SQL/PostgreSQL.pm
jtfrey: I cede all copyright stuff to the project; if accepted, my small contribution will at least be documented here. :-)

jtfrey:
New commit to my development branch this a.m.

1. The globbing in SQL.pm to locate the file associated with the "database driver" name can be avoided by setting "database driver" to the full file system path of the driver file (e.g. "/usr/share/perl5/vendor_perl/Warewulf/DataStore/SQL/MySQL.pm").
2. Minor additions to SQL/PostgreSQL.pm:  database schema in leading comment section now includes a few additional indices as well as a dbvers row in the meta table.
3. Added SQL/SQLite.pm to allow an SQLite db file to be used as the backing store
    - "database name" = file system path of the db file
    - "database user" and "database password" are not used (naturally)
    - default chunk size is 1 MB
    - SQLite3 schema for the database present in leading comment section of the module

gmkurtzer:
The code looks good on this at first glance. I am juggling quite a bit at the moment, so I can't do a full test yet. Has anyone else been able to test this?

Thanks and looks fantastic @jtfrey!

jtfrey:
Understood, @gmkurtzer .  I hope you don't mind my keeping this a moving target, but I keep massaging things.

Latest push to my own development branch:

1. Seems like the "hybrid" approach -- having binstore as files on disk rather than chunk in db table -- might be a good choice for the SQLite implementation.  So rather than having *_hybrid classes, the functionality was merged into the SQL::BaseClass.
2. The "binstore kind" now allows the sysadmin to choose "database" or "filesystem" for object storage.  Each implementation has a default kind ("database" for MySQL/PostgreSQL, "filesystem" for SQLite).
3. Each implementation has a function which returns its schema; the 10-database.init script uses this when setting up the database.
4. All configuration keys are documented in SQL::BaseClass (and implementation-specific keys for each subclass).
5. By default, the entire subsystem will still function the same as it did _before_ all of these changes, including the max_packet_size - 768KB behavior for MySQL chunk size.  And MySQL is still the default database driver.
6. A "database port" configuration key is now present, for anyone running their database on a non-standard TCP/IP port.

The 10-database.init has been rewritten to handle both MySQL and PostgreSQL database setup.  The "database password request" stuff from PR # 14 -- as well as use of database host, port, server -- is respected by the script.  I've done extensive tests on a CentOS 7 machine for both MySQL and PostgreSQL setup from scratch, and it seems to work properly.  E.g. for PostgreSQL with the 'filesystem' binstore:

```
[root@login wwinit]# wwinit database
database:     Checking to see if RPM 'postgresql-server' is installed        OK
database:     Activating Systemd service: postgresql
database:      + /bin/systemctl -q enable postgresql.service                 OK
database:      + /bin/systemctl -q restart postgresql.service                OK
Enter databas Creating database warewulfCREATE DATABASE
                                                                        SUCCESS
database:     Creating database schema                                  SUCCESS
database:     Updating database permissions for wwuser user             SUCCESS
database:     Updating database permissions for wwroot user             SUCCESS
database:     Checking binstore kind                                    SUCCESS
database:     Checking for binstore filesystem path                     SUCCESS
database:     Creating binstore directory /var/lib/warewulf/binstore    SUCCESS
Done.
```

and an extant MySQL:

```
[root@login wwinit]# wwinit DATABASE
database:     Checking to see if RPM 'mysql-server' is installed             NO
database:     Checking to see if RPM 'mariadb-server' is installed           OK
database:     Activating Systemd service: mariadb
database:      + /bin/systemctl -q enable mariadb.service                    OK
database:      + /bin/systemctl -q restart mariadb.service                   OK
Enter databas Database version: 1oot user:database:
database:     Updating database permissions for base user               SUCCESS
database:     Updating database permissions for root user               SUCCESS
Done.
```

The odd "Enter databas" seen there is due to the prompting for the database password.

gmkurtzer:
The moving target is great, plus then I don't feel so bad about pushing off the full review until you feel it is done. :)

Thanks again @jtfrey, this looks sweet and I'm excited to check into it in more details!

bensallen:
Couple comments on packaging and one bug.

1. Should we create sub-packages for warewulf-common-mysql, warewulf-common-pgsql, warewulf-common-sqlite? Otherwise all users are going to need to install perl-DBD-Mysql, perl-DBD-Pg, and perl-DBD-Sqlite. Then only use one.

2. common/lib/Warewulf/DataStore/SQL/Makefile.am: MySQL_hybrid.pm needs to be removed and SQLite.pm added.

macdems:
> Should we create sub-packages for each DB, i.e. warewulf-common-mysql, warewulf-common-pgsql, warewulf-common-sqlite? Otherwise all users are going to need to install perl-DBD-Mysql, perl-DBD-Pg, and perl-DBD-Sqlite. Then only use one.

@bensallen I don't know about RPM. For DEBs it is possible to specify dependencies like `libdbd-mysql-perl | libdbd-sqlite3-perl | libdbd-pg-perl`, which means ‘any of’. However, in such case `wwinit` should verify if the right package is installed.

jtfrey:
> Should we create sub-packages for each DB, i.e. warewulf-common-mysql, warewulf-common-pgsql, warewulf-common-sqlite? Otherwise all users are going to need to install perl-DBD-Mysql, perl-DBD-Pg, and perl-DBD-Sqlite. Then only use one.

@bensallen Separate RPMs for each SQL plugin seems like the logical way to go — ensure necessary dependencies only for the SQL plugin(s) the sysadmin installs.  In which case it also seems to make sense to make that part of `wwinit database` (e.g. `--driver=SQLite` or prompt for selection).

common/etc/functions
```diff
@@ -187,6 +207,10 @@ wwreqroot() {
     fi
 }
 
+wwishost() {
```
bensallen: @jtfrey It appears the function wwishost() was half completed, and isn't actually used anywhere. Can it be removed?

jtfrey:
Ouch...my apologies on that one.  Yes, that function can go away.  I believe I was going to use it to check when the database hostname is the host on which the wwinit command was run.

Issue #16, opened 2017-05-28
GRUB installation problem


macdems:
I have noticed that GRUB installation in `80-mkbootable`  is very badly organized. It is importable and very likely to fail on many systems. My tests (with Ubuntu) have shown that Warewulf does very well for stateless and statelite provisioning, however if fails with stateful one.

I would suggest to refactor this. Here are my suggestions:

1) `grub.cfg`  or `grub.conf` should probably be created per VNFS — it will be much easier to consider differences between different distros and will allow to setup stateful node with different kernel than set in the boostrap (can make sense if the provisioned system is different one than the master one).

2) `grub-install` command should be defined in the database per VNFS. This will make it safe for any future system changes and maybe will even allow to use different bootloader than grub. The safest thing is to define the whole command line with some template for node-specific data (like boot partition).

3) For stateful provisioning Warewulf should not try to call `/sbin/init`. Probably `wwgetvnfs` should not test for it. Instead after provisioning the system should be rebooted and set to booting from disk (for example provision `bootlocal` parameter should allow value `next:xxx` which means: boot wit warewulf, provision and set `bootlocal` to _xxx_.

I suggest to discuss this issue as the refactor would probably require many changes in many places.

bensallen:
Overall 80-mkbootable has seen a lot of evolution across the last few years. It's largely tested on RHEL6 and 7.

1. It probably makes sense to allow a configurable override for specifying where to find a grub.cfg and leave the existing functionality as well. grub2-install is fairly standardized across distros now, so I'm not sure how much need there actually is for this.

2. Configurable override to run a given "WWBOOTINSTALLCMD" makes sense.

3. See provision/initramfs/capabilities/setup-filesystems/99-postreboot. Note, I'm pretty sure /tmp/stateful-root isn't being created anywhere, so this needs to be patched.

All the changes should be within 80-mkbootable. I'm making a few changes in #12 to 80-mkbootable, so it would be good to wait before doing any work until that work is merged.

macdems:
In Ubuntu `/usr/sbin/grub2-install` is renamed to `/usr/sbin/grub2-install`.

jmstover:
@bensallen Hrmm... if `/tmp/stateful-root` isn't being created, should add it into `provision/initramfs/capabilities/setup-filesystems/20-filesystems` at the beginning where it looks for a '/' mountpoint in FILESYSTEMS.

[...]

PR #18, opened 2017-05-29
Add postreboot provision CLI arg, and remove check for /tmp/stateful-root

bensallen:
/tmp/stateful-root isn't being created by 20-filesystems, and I think its sane to reboot no matter what if WWPOSTREBOOT is set. Fixes one of the requests in #16.


gmkurtzer:
Looks good to me too!

PR #20, opened 2017-06-02
mkbootable hack for Debian and Ubuntu.

macdems:
I have added quick hack to make mkbootable work for Ubuntu (and Debian most probably). Refs. #16 .


bensallen:
@macdems Thanks for this. I want todo a rewrite of this script to clean it up in a fairly short time frame, but I'm fine with accepting this for now if @jmstover is happy with it.

One potential fix is the serial console support won't work for Debian/Ubuntu in this PR. The same code that adds values into /etc/default/grub for RHEL should work for Debian/Ubuntu. Also `update-grub` is wrapper for running `grub-mkconfig -o /boot/grub/grub.cfg`.

gmkurtzer:
Hi @macdems,

This looks good to me too, but I have a request... Can you resubmit this PR against the development branch?

Thank you!

macdems:
Closing in favor of #24 and #25 which offer much more elegant solution to #16.

PR #23, opened 2017-06-03
Added some manpages.

macdems:
Added manual pages for some binaries. Originally I made them for the Ubuntu package as required by the policy, but the project may benefit from them in general.


bensallen:
Assuming you agree, please add the full project copyright statement above your own in these files. These look great.

macdems:
Done. The templates were generate automatically. So I did not notice the copyright lines. I removed myself as the manpages are almost exact copies of the `wwsh help xxx` or `wwxxx --help`.

There is one catch though: if anybody ever change anything in the documented commands, the manpages must be manually updated!

PR #24, opened 2017-06-03
Rewrite 80-mkbootable

### LINKED ISSUE #17: 80-mkbootable add EFI Support
After #12 is complete we can provision GPT labeled disks and vfat ESP partitions. We'll need to detect if we're on a machine that was booted via EFI, presumably mount `mount -t efivarfs efivarfs /sys/firmware/efi/efivars` for grub2-install.

bensallen:
## A rewrite of 80-mkbootable to be more flexible.

- Adds support for EFI so this will close #17.
- Deals with distros various paths for grub2-install, grub-install, grub2-mkconfig, grub-mkconfig so should supersede #20.
- Does a reasonable job figuring out where grub.cfg should go, including RHEL/Fedora's pre-installed grubx.efi which expects a config file at a path like /boot/efi/EFI/fedora/grub.cfg. A quick override, if needed within the VNFS, is to create a symlink /etc/grub.cfg or /etc/grub-efi.cfg, like RHEL/Fedora, pointing to the correct location.
- All variables are overridable, but today would be configured via wwsh object. No CLI options.
- Currently this doesn't support Grub 1, which means this won't work with RHEL6 era distros. 
- No longer (re)creates the initrd via dracut or mkinitrd. For now, assuming whatever initrd is in the VNFS is good enough. Lets see if this breaks for anyone.
- Tested on top of #12

### DIFF
--- a/provision/initramfs/capabilities/setup-filesystems/80-mkbootable
+++ b/provision/initramfs/capabilities/setup-filesystems/80-mkbootable
@@ -17,17 +17,17 @@ set -o pipefail
 
 [ -f /etc/functions ] && . /etc/functions
 
-NEWROOT="${NEWROOT:=/newroot}"
+NEWROOT="${NEWROOT:-/newroot}"
 BOOTLOADER="${WWBOOTLOADER:-}"
 
 GRUBINSTALL="${WWGRUBINSTALL:-}"
 GRUBVERS="${WWGRUBVERS:-}"
 GRUBMKCONFIG="${WWGRUBMKCONFIG:-}"
 GRUBCFG="${WWGRUBCFG:-}"
-GRUBDEFAULTCONF="${WWGRUBDEFAULTCONF:=/etc/default/grub}"
+GRUBDEFAULTCONF="${WWGRUBDEFAULTCONF:-/etc/default/grub}"
 
 CONSOLE="${WWCONSOLE:-}"
-KARGS="${WWKARGS:=quiet}"
+KARGS="${WWKARGS:-quiet}"
 
 # Parse the kernel console option for a serial TTY into values useful for /etc/default/grub
 parse_console() {
@@ -36,10 +36,10 @@ parse_console() {
     set -- $(echo "${CONSOLE}" | sed -nr "s/.*ttyS([0-9]),([0-9]{4,6})([n,o,e])?([5-8])?(r)?.*/\1 \2 \3 \4/p")
     IFS="${OIFS}"
 
-    export SERIALUNIT="${$1:-}"
-    export SERIALSPEED="${$2:=115200}"
-    export SERIALPARITY="${$3:=no}"
-    export SERIALWORD="${$4:=8}"
+    export SERIALUNIT="${1:-}"
+    export SERIALSPEED="${2:-115200}"
+    export SERIALPARITY="${3:-no}"
+    export SERIALWORD="${4:-8}"
     
     if [ "${SERIALPARITY}" = "n" ]; then
         export SERIALPARITY=no
--- a/provision/initramfs/capabilities/setup-filesystems/80-mkbootable
+++ b/provision/initramfs/capabilities/setup-filesystems/80-mkbootable
@@ -180,6 +180,29 @@ check_and_mount() {
   fi
 }
 
+# Grub's 10_linux checks for device UUID symlinks exist in /dev/disk/by-uuid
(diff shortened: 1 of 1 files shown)


bensallen:
@macdems When you have a moment could you test this with Ubuntu?

macdems:
Ok, just give me a couple of days.

macdems:
Can you list the config variables and how to change them with `wwsh object`?

bensallen:
@macdems The only setting thats required is the existing `wwsh provision set --bootloader`. 

However the following variables will respect if a value is already set thus allowing override:

GRUBINSTALL - Space delimited list of paths to search for grub-install and/or grub2-install in the VNFS. Defaults to `/usr/sbin/grub2-install /usr/sbin/grub-install /sbin/grub-install`

GRUBMKCONFIG - Space delimited list of paths to search for grub-mkconfig and/or grub2-mkconfig in the VNFS. Defaults to `/usr/sbin/grub2-mkconfig /usr/sbin/grub-mkconfig`
GRUBCFG - Path to output grub.cfg. No default per se, but see find_grub_cfg(). It should work for RHEL and Ubuntu.

GRUBDEFAULTCONF - Path to config file for scripts in /etc/grub.d, defaults to /etc/default/grub.

These can bet set via `wwsh object modify -s GRUBINSTALL=/usr/sbin/grub2-install n01`. Note, any spaces in values need to escaped.

macdems:
I have added a PR #25 agains this branch to make it work with Ubuntu. There were two problems to solve:

1. `grub-install --version` returned nothing if not called from chroot.

2. There was a force reboot without calling `sync`. This resulted with GRUB files not being written to disk. I have added sync command to the `functions` file instead of `80-mkbootable`, as this should resolve any potential problems of this kind with other scripts.

bensallen:
@macdems Both interesting.

1. This seems like a fine workaround, although I'm interested why this is happening. Which version of Ubuntu are you testing on?

2. wwreboot and thus _doreboot is only called when a script fails (exit 2). I agree its a reasonable idea adding a sync into it, but it being called todo a reboot points to something failing in testing.

Also we should split this and another fix I see in wwreboot into its own PR. setsid doesn't know anything about PATH, so cttyhack should be the absolute path /bin/cttyhack, similar to how we're using it in /init.

macdems:
@bensallen 

1. I am testing with Ubuntu Xenial. As you mentioned in your commit, this is probably due to the problems with finding libraries.

2. Indeed, the problem seems to be with the `95-umount`. I don't know why it contains the lines:

   ```
   chroot $NEWROOT /bin/mount -t tmpfs none /dev/shm
   chroot $NEWROOT /bin/mount -t proc none /proc
   chroot $NEWROOT /bin/mount -t sysfs none /sys
   ```
   They all fail, as these directories are already mounted. I don't understand why they are there at all. Shouldn't they be **un**mounted?

   Probably `sync` can be removed from ` _doreboot`, although I think it is reasonable to keep it there. However `95-umount` should be fixed and the reboot should be done by `99-postreboot`. In any case, there should be some error message.

macdems:
I have added the fix in #25. `/newroot/proc` and `/newroot/sys` are properly unmounted, so `95-umount` can succeed.

bensallen:
@macdems I've been testing this PR on-top of #12, which removes these chroot mount lines in 95-umount. I can't find a reason these mounts should exist, especially in a script called 95-umount.

I don't *think* there's a reason we need to umount /sys and /proc in mkbootable. The updated 95-umount in #12 should handle it.

Note a quick way to test #12 + #24 together:

```
git checkout -b testing origin/development
git merge origin/parted
git merge origin/mkbootable
```

macdems:
Ok, it seems the combintation of #12 and #24 does work well. However, actually #12 does not work. The partitions are formatted according to the current `filesystems setting`.

bensallen:
Note this PR is dependent on #12, which is waiting on license sign-off.

jmstover:
Err... didn't that happen?

https://github.com/warewulf/warewulf3/commit/5909c8d9f59f5bd200c9a85ce29e43eac47a56b8

bensallen:
@jmstover Yea, the problem is David doesn't represent the FSF, who is the copyright holder of Parted.

jmstover:
Ahh... Haven't used parted enough to realize it was a full on GNU project.

bensallen:
Merged since we merged #12.

PR #27, opened 2017-06-05
Bootstrap.conf - Add vfat, NLS Codepage 437, and NLS ASCII, Efivarfs, and kernel/drivers/firmware/efi/, and virtio

bensallen:
- Updating the default bootstrap.conf to add vfat, NLS Codepage 437, and NLS ASCII, Efivarfs, and kernel/drivers/firmware/efi/ to add support for EFI ESP (/boot/efi vfat parition for EFI booting) and Efivars. Note, the NLS modules are needed in Debian but are compiled into the RHEL kernel. These are the character encoding schemes for vfat that are default for most distros and are needed for mounting vfat. 

- Virtio is useful for use within virtual machines.


gmkurtzer:
Looks good to me!

PR #28, opened 2017-06-05
Add syncs before reboots, Setsid / cttyhack path fix

bensallen:
Add syncs before reboots. 

Setsid doesn't know about PATH, so cttyhackand sh needs to be absolute paths. We fixed this issue in /init a while back.


gmkurtzer:
I concur.

PR #30, opened 2017-06-06
Limit RPM Upgrade %post wwsh object canonicalize

bensallen:
- Only run object canonicalize on node and file types. These are the only object types that implement canonicalize() beyond a no-op. Currently every object thats being iterated is running persist() triggering *.modify events in Warewulf/DataStore/SQL/MySQL.pm. This causes all bootstraps to be rebuilt needlessly.

- Include a note to users on upgrade of warewulf-provision rpm to run bootstrap rebuild.


gmkurtzer:
I concur, but you guys beat me to it! ;-)

PR #32, opened 2017-06-07
Fixed warewulf-common test.

macdems:
Right now the test suite fails due to the added GITVERSION to `wwconfig`. This is a simple fix to ignore it (it may be empty). By default all tests must pass to build deb package.

### DIFF
--- a/common/t/10-acvars.t
+++ b/common/t/10-acvars.t
@@ -38,8 +38,8 @@ plan("tests" => (
 $acvars = new_ok("Warewulf::ACVars");
 can_ok($acvars, "new", "get", "vars");
 
-# Make sure we account for all the variables that exist currently.
-cmp_ok(scalar(@var_names), '==', scalar($acvars->vars()),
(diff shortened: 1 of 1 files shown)


jtfrey:
Might be better to explicitly check each necessary ACVar key for a value and then do the cmp_ok() against that, e.g.

```
# Make sure we account for all the variables that MUST have a
# value:
my $count = 0;
foreach my $var_name (@var_names) {
    $count++ if defined($acvars->get($var_name));
}
cmp_ok(scalar(@var_names), '==', $count,
       "All Warewulf::ACVars variable names accounted for");
```

I've implemented this in my development branch, FWIW.

gmkurtzer:
Agreed @jtfrey. @macdems, do you concur?

macdems:
Putting GITVERSION in `@var_names` breaks sanity checks in lines 80-90. So it must be handled individually.

macdems:
I don't like what @jtfrey proposed. It's better to check for all the variables and then explicitly allot GITVERSION to be empty or undefined.

PR #33, opened 2017-06-08
Use devtmpfs if available to populate /dev instead of mdev

bensallen:
Warewulf's existing implementation uses mdev. Mdev uses the kernel's asynchronous hotplug interface to be executed to populate /dev.

The current approach can cause a race condition on partitioning local disk, the devices being created/updated and formatting (or whatever the next step to access that partition's device). Devtmpfs has been around in the kernel at least since 2.6.32, so it's available in CentOS/RHEL 6.


bensallen:
Merging this since I've been using it in production for a few weeks now without issue.

PR #37, opened 2017-06-09
Optionally Validate VNFS Checksum On Boot

bensallen:
Add functionality to validate VNFS checksum on boot.

- Add new --validate boolean CLI arg for VNFS checksum validation
- Use tee to duplicate fifos so both md5sum and gunzip can consume the vnfs stream from wget
- Adds new variable WWVALIDATE_VNFS in the initrd to enable/disable md5sum validation
- Adds more feedback to console and syslog on each step of wwgetvnfs. Will display checksum fetched from the master even if WWVALIDATE_VNFS isn't enabled.
- Add VNFS_NAME and VNFS_CHECKSUM to variables exported for nodes in nodeconfig.pl
- Cleanup quoting and syntax to make shellcheck linter happy in shell scripts.

### DIFF
--- a/common/share/wwsh.1
+++ b/common/share/wwsh.1
@@ -704,6 +704,11 @@ Define the bootstrap image that this node should use.
 Define the VNFS that this node should use.
 .RE
 .PP
+.B \-\-validate        
+.RS 4
+Enable checksum validation of VNFS on boot.
+.RE
+.PP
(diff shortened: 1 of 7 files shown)


bensallen:
Merging this as I've been using in my production environment for a few weeks successfully, and it defaults to existing behavior.

PR #41, opened 2017-06-15
Remove conditional that filters out duplicate array members

### LINKED ISSUE #38: Cannot have Duplicate Lines in Array Objects
#12 needs to be able to have duplicate array entries like `mktable gpt`. The code at https://github.com/warewulf/warewulf3/blob/master/common/lib/Warewulf/Object.pm#L252 appears to deduplicate array entries. Can the conditional on line 252, `if (!scalar(grep { $_ eq $newval } @{$self->{$key}}))` be removed? What was the original intention for this?

bensallen:
Fixes #38.

### DIFF
--- a/common/lib/Warewulf/Object.pm
+++ b/common/lib/Warewulf/Object.pm
@@ -248,10 +248,7 @@ add()
     }
     foreach my $newval (@vals) {
         if (defined($newval)) {
-            # NOTE:  This prevents undef from being a value in this array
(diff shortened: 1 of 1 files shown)


bensallen:
Going to merge this since its easy to revert, so it gets some testing. Let me know if it is a terrible idea.

gmkurtzer:
I concur. Thanks!

Issue #35, opened 2017-06-08
ARM64 Boot


ericvh:
So, as part of OpenHPC we've been trying to use warewulf to provision ARM64 servers which are UEFI based.  Right now the work-around is to netboot grub and have a grub.cfg contain the provision and boot information for the nodes.  The current solution is documented (somewhat) here: https://github.com/openhpc/ohpc/wiki/ARM-Tech-Preview

I was somewhat hoping changes to grub2 and the UEFI boot would magically fix my problems, but it doesn't seem there is a clear path to fixing the work around (ie. no ARM support on the horizon for SYSLINUX and grub2 syslinux_confligfile doesn't seem to "do the right thing").  

The problems I am currently running into:

a) No syslinux equivilent for ARM or ARM64
b) Current ARM server systems need some extra bootloader support for UEFI/FDT embedding so you need to boot Image files versus raw kernel.
c) grub appears necessary as a step before (b) to make sure FDT is loaded appropriately

I'm willing to accept that I'm doing something stupid and there is a solution to one or all of the above.

Here's the alternatives I'm currently considering:

a) Modify warewulf to provision ARM nodes purely through grub netboot and grub.cfg instead of using syslinux formats.  This is the most straightforward IMHO, but may require a slightly different set of parameters for ARM versus x86 and would require embedded grub2 in a similar fashion to the current embedding of syslinux.

b) Modify grub2 to fix syslinux_configfile to work as expected - not sure how well this will go over in the grub2 folks.  The other problem of needing to load Image versus "kernel" will remain, but maybe that's a more straightforward fix?  In this mode, grub2 could be used as a "drop-in" for syslinux.

c) Further modification of grub2 to be able to more efficiently handle raw kernel files versus Image files.

Ideas?  Happy to jump on slack/irc/whatever to discuss if this community uses that sort of thing.

bensallen:
Hi Eric,

This is a topic I'm also actively looking into. Grub2 does appear to be the only UEFI aarch64 network capable boot loader available today. 

One other possible item is using an UEFI implementation that has a ramdisk and TFTP or HTTP clients. Copying the kernel and initrd into a ramdisk, and booting the it via EFI stub. This approach provides a number of challenges. Relying on the vendor to implement said functionality in their UEFI. These functions don't appear to be standard in EDKII. We would also need to generate a EFI shell script to be the pxelinux.cfg equivalent. HPE is the only vendor that I know of that can pull this off today, and only with x86_64 hardware [1].

[...]

PR #45, opened 2017-07-20
Install bootloader and initramfs components into a new architecture specific subdirectory

bensallen:
syslinux (for now), the base rootfs and capability files are now installed under architectural specific directories, eg:

/usr/share/warewulf/x86_64
/var/warewulf/initramfs/x86_64

Various bits of code need to be updated to use this new path, but that will come as a separate PR. 

This is part of the effort to support ARM64: #35.


bensallen:
Going to combine this and other ARM64 work into #44 and rename it.

PR #46, opened 2017-07-30
Replace Syslinux with iPXE as Network Bootloader

bensallen:
- Replace Syslinux with iPXE as Network Bootloader
a. iPXE script generation per host stored at /var/lib/tftpboot/warewulf/ipxe/cfg/<hwaddr>
b. iPXE bootloader files installed under /var/lib/tftpboot/warewulf/ipxe/bin-*

- Cross-compile support of iPXE for aarch64 and x86_64
a. provision's ./configure now has a --enable-cross-compile
b. provision's RPM spec file now has a cross_compile define.

This PR should be merged after #44.

iPXE is GPLv2 code, and requires we ask the maintainers to sign-off that Warewulf isn't a derivative work.


bensallen:
A note for whoever ends up merging this, this will conflict with #44, the files in this PR should be used as-is, i.e.

```
git checkout --theirs -- 3rd_party/Makefile.am
git checkout --theirs -- etc/dhcpd-template.conf
git checkout --theirs -- lib/Warewulf/Provision/Pxelinux.pm
git checkout --theirs -- warewulf-provision.spec.in
```

PR #47, opened 2017-07-30
Approve distribution of iPXE within Warewulf

mcb30:
As per discussion via e-mail, the distribution of iPXE as a standalone
component within Warewulf qualifies as "mere aggregation" under the
terms of the GPL and is in conformance with the licensing terms.

This is not a copyright waiver and does not modify the licensing terms
applicable to iPXE.

Signed-off-by: Michael Brown <mcb30@ipxe.org>


bensallen:
Thanks Michael.

PR #61, opened 2017-08-16
Set MySQL permissions based on client hostname. Fix a few typos in MySQL grant statements.

bensallen:
- Set MySQL permissions based on client hostname. 
- Fix a few typos in MySQL grant statements, transposing DBROOTUSER and DBUSER
- Grant `SELECT, INSERT, UPDATE, DELETE` to DBROOTUSER.

### DIFF
--- a/common/libexec/wwinit/10-database.init
+++ b/common/libexec/wwinit/10-database.init
@@ -176,7 +176,8 @@ if [ "$DATASTORE" = "mysql" ]; then
     elif [ -n "$DBSERVER" ]; then
         DBCLIENT=$(hostname -f)
     else
-        wwprint "Database server undefined, so can't determine what database permissions should be for this client"
+        wwprint "Database server undefined! Can't determine what database permissions should be for this client\n" RED
+        exit 255
     fi 
 
     if [ -n "$DBUSER" ] && [ "$DBUSER" != "root" ]; then
--- a/common/libexec/wwinit/10-database.init
+++ b/common/libexec/wwinit/10-database.init
@@ -162,7 +162,7 @@ if [ "$DATASTORE" = "mysql" ]; then
     else
         wwprint "Database version: UNDEF (need to create database)\n"
         perl -e 'use Warewulf::DataStore::SQL::MySQL;print Warewulf::DataStore::SQL::MySQL->database_schema_string();' > "${SCRIPT_TMPDIR}/schema"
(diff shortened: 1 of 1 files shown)


common/libexec/wwinit/10-database.init
```diff
@@ -171,20 +171,28 @@ if [ "$DATASTORE" = "mysql" ]; then
         fi
     fi
 
-    if [ -n "$DBUSER" -a "$DBUSER" != "root" ]; then
+    if [ -n "$DBSERVER" ] && ([ "$DBSERVER" == "localhost" ] || [ "$DBSERVER" == "127.0.0.1" ]); then
+        DBCLIENT="localhost" 
+    elif [ -n "$DBSERVER" ]; then
+        DBCLIENT=$(hostname -f)
+    else
+        wwprint "Database server undefined, so can't determine what database permissions should be for this client"
```
rengolin: Shouldn't this exit, too?

bensallen: Yes it should, thanks.

rengolin:
LGTM. Thanks!

bensallen:
The root user is evidently now defaulted to 'wwroot', so yes we do.

jmstover:
Ugh... so then it's not the root user, but a different user with more permissions on the DB. Got ya.... normal user only has SELECT.

This is so going to cause issues external scripts. You can't read the database-root.conf file unless you are root, and scripts are expected to run as a normal user to do object manipulation, so expect DBUSER to have access to do that. :P

bensallen:
@jmstover 

If DBROOTUSER or DBUSER is 'root' we no-op the GRANTs, so a user can certainly still use root, but this isn't best practice in general.

The limitations you laid out I see as functioning as intended. I don't want the user in database.conf to have write privileges, *unless* database-root.conf doesn't exist to mirror old behavior. However, we seem to be effectively requiring database-root.conf now as of #15 in this wwinit, so at least with wwinit we're not mirror old behavior around only database.conf existing with wwinit now.

Since our permissions model is so limited, I'm not sure what else we should do here. As reasonably "secure" defaults, I *think* this is best we can do.

jmstover:
For #15... I missed the continuation of that, got filtered out of my inbox it looks like, and completely forgot its existence. Wanted to test that too, before I got buried here at work. :/

I'll just need to remember the change, as I have scripts that can't read database-root.conf (normal user) but need to make changes to the data which is feasible in the SVN version I'm running. But wouldn't be with these changes.

PR #62, opened 2017-08-16
wwbootstrap: Attempt to gunzip the kernel, otherwise copy as previous behavior

bensallen:
- Attempt to gunzip the kernel, aarch64 kernels are compressed and iPXE can't boot gzip compressed kernels.
- If the kernel isn't a gzip, IO::Uncompress::Gunzip makes a direct copy of the file.

### DIFF
--- a/vnfs/bin/wwbootstrap
+++ b/vnfs/bin/wwbootstrap
@@ -16,7 +16,7 @@ use Getopt::Long;
 use File::Path;
 use File::Basename;
 use File::Copy;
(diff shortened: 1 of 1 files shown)


rengolin:
This commit looks uncontroversial to me.

jmstover:
Almost got to it before merge... but I agree it looks fine. Just saying. ;)

bensallen:
Thanks @jmstover

PR #63, opened 2017-08-16
Serve the contents of the tftpboot directory via httpd

bensallen:
Allows iPXE to fetch the kernel and initrd.gz via HTTP instead of TFTP.


rengolin:
This commit looks uncontroversial to me.

PR #67, opened 2017-08-24
Skip updating pxe cfg for nodes that don't have a bootstrap defined

bensallen:


bensallen:
@jmstover I'm not quite sure how I ended up here... 

I have a few hosts defined in Warewulf just so they get DHCP reservations. They've never had bootstraps or VNFS defined (I think), but when I go and dump one of the hosts the bootstrap and VNFS ID is set. 

Ran into this issue after upgrading from a dev version after the parted and mkbootable work, to current dev as of today. This particular install has been around since 3.6 however.

If I go set `sudo wwsh provision set --vnfs=UNDEF --bootstrap=UNDEF` the host, their VNFS and Bootstrap entries do go away, so I'm going to caulk this up to my install getting cruddy. 

I still think this PR still is a reasonable check however.

bensallen:
Maybe we should change the warning message a touch, to something like `Bootstrap defined for $nodename, but bootstrap doesn't exit, skipping...`

jmstover:
That sounds perfect to me.

bensallen:
Also changed it to a debug message instead of warning to match the other bootstrap doesn't exist message a few lines down.

PR #69, opened 2017-08-29
Avoid uninitialized warning when $arch is clearly not defined

rengolin:


jmstover:
@bensallen I'll leave you to merge when you're ready since you've had more of a hand in this than I, but I don't see an issue with the move here.

PR #81, opened 2017-10-10
Remove generation of SSHv1 host keys

jabl:
The SSHv1 protocol has been replaced by SSHv2 a long time ago, and
openssh has recently completely removed SSHv1 support.

### DIFF
--- a/vnfs/etc/wwmkchroot.tmpl
+++ b/vnfs/etc/wwmkchroot.tmpl
@@ -79,6 +79,9 @@ configure_fstab() {
 
 configure_sshkeys() {
     echo "Creating SSH host keys"
+    # SSHv1 is old and busted, but if you need it, uncomment this to
+    # generate a SSHv1 RSA host key
+    #/usr/bin/ssh-keygen -q -t rsa1 -f $VNFSDIR/etc/ssh/ssh_host_key -C '' -N ''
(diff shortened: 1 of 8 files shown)


vnfs/etc/wwmkchroot.tmpl
```diff
@@ -79,7 +79,6 @@ configure_fstab() {
 
 configure_sshkeys() {
     echo "Creating SSH host keys"
-    /usr/bin/ssh-keygen -q -t rsa1 -f $VNFSDIR/etc/ssh/ssh_host_key -C '' -N ''
```
jmstover: Can you comment this out, instead of removing?

bensallen:
These changes look good to me.

jabl:
Thanks for the prompt review.

So is it good as is, or should I comment it out instead of removing? And if I comment it out, is it sufficient to comment it out in wwmkchroot.tmpl or should it be done in all files touched by this patch?

FWIW, I just checked, and on a CentOS 5 (released 2007, EOL 2017) machine, the default installation has created a SSHv1 host key. On CentOS 6 (released 2010, EOL 2020) a SSHv1 host key is no longer created.

jmstover:
Just commenting out of the tmpl file is good. Just so there's at least 1 place the rsa1 key generation is at.

And yeah... I've heard of places still running 5 because they're running old software that hasn't been updated in ages. So... just to have that small part mentioned in the there still just in case it's needed...

jmstover:
Excellent. Thanks!

PR #83, opened 2017-10-11
Generate RSA ssh keys for cluster access instead of DSA

jabl:
OpenSSH 7.0 disabled support for DSA keys by default. While support
can be reenabled via configuration, they were removed due to security
concerns, so it seems safer to switch to another algorithm. Ed25519
might in some sense be the best, but it's quite new and might thus not
yet be available on all distros Warewulf supports. So use RSA which is
good enough for now.

http://www.openssh.com/txt/release-7.0
http://www.openssh.com/legacy.html
https://meyering.net/nuke-your-DSA-keys/
https://security.stackexchange.com/questions/5096/rsa-vs-dsa-for-ssh-authentication-keys

### DIFF
--- a/cluster/bin/cluster-env
+++ b/cluster/bin/cluster-env
@@ -17,7 +17,7 @@ fi
 if [ ! -f "$HOME/.ssh/config" -a ! -f "$HOME/.ssh/cluster" ]; then
     echo "Configuring SSH for cluster access"
     install -d -m 700 $HOME/.ssh
-    ssh-keygen -t dsa -f $HOME/.ssh/cluster -N '' -C "Warewulf Cluster key" > /dev/null 2>&1
+    ssh-keygen -t rsa -f $HOME/.ssh/cluster -N '' -C "Warewulf Cluster key" > /dev/null 2>&1
     cat $HOME/.ssh/cluster.pub >> $HOME/.ssh/authorized_keys
     chmod 0600 $HOME/.ssh/authorized_keys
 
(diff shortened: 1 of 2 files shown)


jmstover:
DSA and ECDSA are the two allowed under the NIST standard (FIPS 186-2). If you go by an updated FIPS 186-4, DSA keys can be 1024, 2048, or 3072 in size.

It looks like they (OpenSSH) implemented DSA when 1024 was the *only* key size allowed, and didn't update ssh-keygen when the standard was updated.

I'm going to have to add @gmkurtzer here to look at this and sign off on it. Because this would have a larger effect.

jabl:
I'm not an expert on US government standards, but it seems to me that per FIPS 186-4 ( http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf ) DSA (chapter 4), ECDSA (chapter 6), and RSA (chapter 5) are allowed. For RSA, the same key sizes as for DSA: 1024, 2048 and 3072 bits (by default current versions of OpenSSH ssh-keygen generate 2048 bit RSA keys).

jmstover:
Okay, scratch what I said. They used ANSI X9.31 in the Federal Register notice, and I didn't put that together with RSA.

>  ANSI X9.31, Digital Signature Using Reversible Public Key Cryptography

https://csrc.nist.gov/News/2000/Announcing-Approval-of-FIPS-186-2,-Digital-Signatu

PR #87, opened 2017-10-17
Modernize filesystem partitioning examples to use ext4

jabl:
Use ext4 in various examples instead of ext2 or ext3. Also mention
that vfat is a supported partition type.


jabl:
Ah, I missed it has been redone for development. Looking at it, you mean provision/initramfs/capabilities/setup-filesystems/20-filesystems, right? Though I'm a bit confused how it works, or how does $WWFS get set to some non-default value? Is there some magic that make it take the value of the FILESYSTEMS object like in https://warewulf.github.io/warewulf3/recipes/stateful-provisioning.html? I.e. that document is correct except the syntax for FILESYSTEMS has changed?

Anyway, thanks for the quick review and merge!

bensallen:
@jabl See here for example input using the development branch: https://github.com/warewulf/warewulf3/tree/development/provision/etc/filesystem/examples, eg.

cp /etc/warewulf/filesystem/examples/gpt_example.cmds /etc/warewulf/filesystem/my_gpt_layout.cmds
wwsh provision set --filesystem=my_gpt_layout n01

The example in 20-filesystems itself is the format on how the configuration is stored and shared with the node, i.e. if you were working by hand with `wwsh object`.

PR #88, opened 2017-11-03
Use ip instead of ifconfig and route

jabl:


jabl:
With this patch it's now possible to not include the net-tools package in the default chroot's at least on RHEL-7 (and derivatives) and Ubuntu 16.04. Do you want me to push such a change as well?

(On RHEL 5/6 net-tools includes tools such as hostname so we can't get rid of it there.)

provision/initramfs/init
```diff
@@ -215,14 +215,17 @@ ifup() {
...
 }
 
 
 # Bring up networks
 msg_white "Bringing up local loopback network:"
-if ifconfig lo 127.0.0.1 netmask 255.255.255.0 up >/dev/null 2>&1; then
+if ip link set dev lo up >/dev/null 2>&1; then
+    # If this addr already exists for this interface ip will return
+    # error, don't fail due to that, just ignore return value.
+    ip address add 127.0.0.1/8 dev lo >/dev/null 2>&1
```
bensallen: Can you tack a `|| true` on this line since we don't care if it fails. While we're not using `set -o errexit`, this does ensure if we ever get to that point, it will be safely ignored.

bensallen:
Hi @jabl, how do these patches have an effect on needing `ifconfig` in the VNFS?

jabl:
After this patch, "git grep ifconfig" in the repo turns up empty. And since e.g. rhel-7 doesn't install net-tools by default any more, the distro init stuff doesn't need it anymore either. So AFAICS the net-tools package (which is the one which contains ifconfig and route) is no longer needed. Haven't tested it, though, so I could be missing something...

bensallen:
@jabl Ah ok, these changes are all in the initramfs using Busybox's `ifconfig`. None of these uses of `ifconfig` run within the chroot of the extracted VNFS, so I don't think net-tools was actually required before these changes. That being said, these are all good changes, modernizing the initramfs. Looks good to me.

PR #89, opened 2017-11-09
Set broadcast address as well when setting ip

jabl:
When using "ip address" to add an address to an interface, the
broadcast address needs to be explicitly set as well.

### DIFF
--- a/provision/initramfs/dhcp-script
+++ b/provision/initramfs/dhcp-script
@@ -11,7 +11,7 @@
 > /etc/network.conf
 
 if test -n "$ip"; then
-   ip address add $ip/$subnet dev $interface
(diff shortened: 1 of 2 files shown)


bensallen:
ip sure does have a goofy syntax for setting the broadcast. What problem were you seeing without it specified?

jabl:
Well, the "+" just means to set all the bits in the host part of the address to 1, which is usually (almost always?) what you want (and what ifconfig does by default).

So I haven't seen a specific problem, I just wanted to make sure it works the same was it used to work before when it was using ifconfig rather than ip. Maybe broadcasts don't work if a broadcast address isn't defined? If you check the network settings on any random Linux machine with "ip a" you'll see that except for the loopback interface, the broadcast address tends to be specified for every address.

Issue #84, opened 2017-10-11
Consider using snponly.efi instead of ipxe.efi for UEFI boot


jabl:
ipxe.efi uses the iPXE network card drivers; at least Dell PowerEdge Haswell NIC's don't work correctly with it in our testing. snponly.efi is the UEFI equivalent of undionly.kpxe for legacy BIOS, and uses the builtin NIC driver and replaces only the "upper" part of the PXE driver stack. Which seems to work.

bensallen:
Did you try testing with snp.efi? It seems like a more universal build compared to snponly.efi.

```
snp same as snponly but tries to boot all devices and not just the one it was chained via, this is also included in ipxe builds
```

Ref: http://ipxe.org/appnote/buildtargets

jabl:
Hmm, sounds reasonable; I wasn't aware of snp.efi. I'll test it and report back (probably won't have time until next week, unfortunately).

jabl:
Yes, I confirm it works with snp.efi as well.

For completess, the product name of the tested system was Dell PowerEdge C6320, and lspci tells that the onboard NIC's are of type:

Intel Corporation 82599ES 10-Gigabit SFI/SFP+ Network Connection (rev 01)

bensallen:
@jabl Thanks for testing. I'll update this when I find some time.

bensallen:
Fixed in #90.

yiya1989:
Is the snp.efi more similar to undionly.kpxe? why not the snponly.efi ? When I use the ipxe, i mostly what to  boot the specific NIC device it was chained from. @bensallen

bensallen:
@yiya1989 See http://ipxe.org/appnote/buildtargets. "snp same as snponly but tries to boot all devices and not just the one it was chained via, this is also included in ipxe builds"

This logic matches our approach in the Warewulf initramfs so it was preferred.

PR #90, opened 2017-11-09
Use snp.efi instead of ipxe.efi for i386 and x86_64 UEFI boot

jabl:
This should fix https://github.com/warewulf/warewulf3/issues/84


bensallen:
Looks good thanks.

Issue #78, opened 2017-09-18
Permissions/ownership update on File objects


jtfrey:
What was the reason for the File `sync()` function only updating the checksum and size?  Why are ownerships and privileges not updated to match the origin file?  If I modify either of those attributes on the origin file, I must remove the file object (and silently have provisioning references to it removed, too) then re-add it, then re-add it to any provisioning profiles on which it had been present.

FWIW, the `sync()` code also doesn't check to ensure that the file type from the datastore still matches the type of the origin file.  It blindly assumes that to be the case and proceeds with opening and reading the file, etc.

jmstover:
Well, at least in our setup, that would break things horribly. We'll get multiple VNFS's that use the same files. But, because it wasn't watched when creating the VNFS, UID's will be all off.

Soooo.... we create multiple files that use the same ORIGIN (as the same file is needed), and just have different `--uid`, `--gid`, and/or `--mode` flags (`wwsh file` command) as needed. 

>  If I modify either of those attributes

Use `wwsh file` ... for example:

  `wwsh file set [Name] --uid=1234 --gid=4321 --mode=0660`

> the `sync()` code also doesn't check to ensure that the file type from 
> the datastore still matches the type of the origin file

Err... why would it? It's a different file at that point. IMO, trying to protect for every eventuality that _could_ happen, is just going to cause a headache. There's way to many of them. It sets on import, and trusts the admin to not do something... strange.

jtfrey:
> Soooo.... we create multiple files that use the same ORIGIN (as the same file is needed), and just have different --uid, --gid, and/or --mode flags (wwsh file command) as needed.

Okay, that was a use case I hadn't considered.  But perhaps there's a way we could augment the existing functionality to make this more straightforward for locales which implement universal uid/gid (including one-off clusters)?

- If `--uid`, `--gid`, or `--mode` options are specified without an accompanying value, `stat()` the origin entity and use its current value for that option.
- Alter the `sync` subcommand to accept the `--uid`, `--gid`, and `--mode` flags (behaving the same way as above) to enable full sync'ing of the file object (content + meta data) against the origin entity. 

> Err... why would it? It's a different file at that point. IMO, trying to protect for every eventuality that could happen, is just going to cause a headache. There's way to many of them. It sets on import, and trusts the admin to not do something... strange.

[...]

PR #110, opened 2018-03-16
Implementation for Issue #78

jtfrey:
Implementation of mode/uid/gid being taken from origin filesystem entity for "file sync" and "file set" subcommands.  E.g.

```
$ wwsh file import /etc/slurm.conf
$ chown slurm:slurm /etc/slurm.conf
$ chmod 0660 /etc/slurm.conf
$ wwsh file set --uid= --gid= --mode= slurm.conf
About to apply 3 action(s) to 1 file(s):

     SET: slurm.conf SET MODE = 0660
     SET: slurm.conf SET UID = 501
     SET: slurm.conf SET GID = 501

Proceed?
Yes/No [no]>
```

The same flags when passed to the "file sync" command updates the datastore content for the file object and also sets those three metadata components.

### DIFF
--- a/common/lib/Warewulf/Module/Cli/File.pm
+++ b/common/lib/Warewulf/Module/Cli/File.pm
@@ -21,6 +21,7 @@ use File::Path;
 use Text::ParseWords;
 use Digest::MD5 qw(md5_hex);
 use POSIX;
+use Fcntl ':mode';
 
 our @ISA = ('Warewulf::Module::Cli');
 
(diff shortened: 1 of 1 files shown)


jtfrey:
Disregard...further testing shows that "reimport" is probably a better approach.

holmanbph:
@jtfrey so do you think that #78 should be closed then?  

Otherwise I have an implementation that would use the syntax `wwsh file sync --overwrite  [file obj]` to indicate that you want gid / uid / mode to be pulled in from the source file (if the overwrite flag isn't given, sync works as it currently does).  

I stopped working on it when you submitted this pull request, but I think it was ready for review.  It just adds a new flag for this functionality so it doesn't break anybody else's existing use cases.

jtfrey:
Yes, I think the --overwrite flag mentioned by @jmstover is a better route to take.

PR #112, opened 2018-04-05
Issue #78 Impementation - auto-sync gid/uid/mode metadata

holmanbph:
Adds new `--overwrite` flag on the file object in `wwsh`.  This flag allows the file object to automatically use the file's metadata for updating the gid / uid / mode on the file object.  

**Hi everyone!  This is my first pull request on this project.  I'm pretty new to Perl in general, as well as Warewulf3, so I'm very open to feedback.  I hope this helps.**

### DIFF
--- a/common/lib/Warewulf/File.pm
+++ b/common/lib/Warewulf/File.pm
@@ -427,7 +427,6 @@ sync()
         if ( $was_stored ) {
             $self->checksum($digest->hexdigest());
             $self->size($total_len);
-            $db->persist($self);
             if ( $metadata ) {
                 if( @statinfo && ( scalar($self->origin())  == 1 ) ) {
                     $self->gid( $statinfo[5] );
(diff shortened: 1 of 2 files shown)


common/lib/Warewulf/File.pm
```diff
@@ -427,6 +428,15 @@ sync()
             $self->checksum($digest->hexdigest());
             $self->size($total_len);
             $db->persist($self);
+            if ( $metadata ) {
```
jmstover: Can you move the  `$db->persist($self);` call _below_ the entire `if ( $metadata ) { ... }` block?

`$db->persist()` is what stores the information back into the database, otherwise you're just changing the information on the working object.

holmanbph: Good catch.  I will do that.

holmanbph:
I made the change required to make the metadata persist.

PR #116, opened 2018-04-12
throttled_sleep debug

jmstover:
When WWDEBUG is set, change some values used in `throttled_sleep()`

### DIFF
--- a/provision/initramfs/capabilities/transport-http/functions
+++ b/provision/initramfs/capabilities/transport-http/functions
@@ -15,10 +15,22 @@ if [ -f "/warewulf/config" ]; then
     . /warewulf/config
 fi
 
+if [ -f "/etc/functions" ]; then
+    . /etc/functions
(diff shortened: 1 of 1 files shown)


provision/initramfs/capabilities/transport-http/functions
```diff
@@ -15,10 +15,22 @@ if [ -f "/warewulf/config" ]; then
...
+    . /etc/functions
+fi
+
 START=0
 BACKOFF=2
 INTERVALS=30
 COUNT=1
+
+# If we want to debug, change some values
+if [! -z "$WWDEBUG" ]; then
```
bensallen: Isn’t ‘-n’ the negate of -z?

jmstover: Yeah... habbit of just doing 'not'. I'll change it to `-n`

PR #122, opened 2018-05-25
Fix issues compiling BusyBox 1.26.2 under RHEL/CentOS 6

DLAtWork:
This PR consists of two commits made against the development branch.

Commit 7e83e0d just does a little cleanup in provision/configure.ac so the AM_CONDITIONAL ISDEB check finishes out with a AS_MSG_RESULT.

Commit e352599 is the real meat of this one. The BusyBox 1.26.2 tar ball included with Warewulf won't compile under RHEL/CentOS 6.9 without patching the source and changing some config options.
Specifically this commit fixes:

1. MTD_MODE_RAW in the mtd_file_modes enum in mtd/mtd-abi.h changed to MTD_FILE_MODE_RAW in kernels after 2.6.32. This commit adds a patch to BusyBox that #define's MTD_FILE_MODE_RAW to MTD_MODE_RAW in nandwrite.c.

2. Some of the options in BusyBox.config need to be turned off so this commit moves the busybox.config -> .config copy line above the loop that patches the BusyBox source in provision/initramfs/Makefile.am.

3. Kernel 2.6.32 doesnt' have BLKSECDISCARD implemented in linux/fs.h so this commit includes a patch to set CONFIG_BLKDISCARD=n in the BusyBox source .config file. 

4. Kernel 2.6.32 doesn't have setns implemented in bits/sched.h so this commit includes a patch to set CONFIG_NSENTER=n in the BusyBox source .config file.

5. Kernel 2.6.32 doesn't have fsync implemented in unistd.h so this commit includes a patch to set CONFIG_SYNC=n in the BusyBox source .config file.

6. This commit updates provision/configure.ac with the AC_CHECK_DECL lines to implement 1 and 3 above.

7. This commit updates provision/configure.ac with the AC_CHECK_FUNC lines to implement 4 and 5 above.

8. Added the new BusyBox patch files to EXTRA_DIST in provision/initramfs/Makefile.am so they would be included in dist-* archives and any SRPMS made from the source files.

The checks in 6 and 7 above are done on declarations and functions so technically these patches would be applied for any system missing these bits, not just RHEL/CentOS 6.

All of these changes applied and compiled cleanly in mock EPEL 6 and 7 x86_64 environments.

The resulting RPMs installed cleanly on CentOS 6.9.


bensallen:
Hi Donald,

Thanks for these. 

Could the .busybox-1.26.2/.config patches be applied to the busybox.config we carry in the repo: https://github.com/warewulf/warewulf3/blob/development/provision/initramfs/busybox.config. Instead of separate patch files?

For the nandwrite patch, is there any signs of this being upstreamed?

DLAtWork:
The .config changes could be put in busybox.config, with the caveat they will disable the BLKDISARD, network namespace, and file sync features of BusyBox on all platforms, even those that would support them like CentOS 7. I made them patches in this pull request so they would only apply to platforms missing those features. If those features don't have any impact on Warewulf, I agree it would be much simpler to update the busybox.config.

It looks like the BusyBox devs recommend using a newer version when you find a bug and aren't keen on backporting fixes for old versions: [https://busybox.net/FAQ.html#bugs](https://busybox.net/FAQ.html#bugs) which is understandable.

I took a quick grep through the 1.28.4 code and their BugZilla for MTD_FILE references. It doesn't look like this IOCTL naming change has been addressed in the latest versions. Would you prefer this patch be submitted to upstream? I'm guessing they'd want the patch against 1.28.4 rather than 1.26.2 based on their FAQ.

Alternatively the "CONFIG_NANDWRITE" and "CONFIG_NANDDUMP" features could be set to "n" in the busybox.config file and the issue would be avoided during compilation. Again, I'm not sure if Warewulf needs these features in BusyBox. ( I have to admit I got over-zealous on this fix since it seemed so simple, rather than just turning off the features, like the other issues. :) )

bensallen:
We can upgrade to the latest Busybox version. We try to reasonably keep up with our 3rd party software versions.

We're not using the commands:

blkdiscard
nsenter
nandwrite
nanddump

CONFIG_SYNC gives me pause, however. We're using the sync command quite a bit across the inittramfs. I setting this to no is actually causing it to not be built, I expect this to break the initramfs. On CentOS 6.9 or otherwise.

DLAtWork:
OK, did some more digging and a quick test.

It looks like setting "CONFIG_FEATURE_SYNC_FANCY" to n would work around this issue as well since that removes the requirement for syncfs() support per this commit: [https://git.busybox.net/buildroot/commit/?id=d42ca918cc72bb1baf817ee1614d77b5081a7e75](https://git.busybox.net/buildroot/commit/?id=d42ca918cc72bb1baf817ee1614d77b5081a7e75)

On my CentOS 6 box with NANDWRITE, NANDDUMP, NSENTER, BLKDISCARD, and FEATURE_SYNC_FANCY disabled, but with SYNC enabled, busybox 1.26.2 compiled successfully.

bensallen:
This approach seems pretty straightforward then. Happy to merge a PR that makes the busybox.config changes.

DLAtWork:
Awesome! Updated the PR with a new commit.

bensallen:
Thanks!

PR #126, opened 2018-09-11
wwinit: Check if service is enabled before enabling it

jjolly:
This avoids ugly error messages.


jmstover:
Thanks

PR #130, opened 2018-09-11
Suse prov busybox use libtirpc

jjolly:
busybox: Newer versions of glibc do not ship rpc functions any more
    
    Use tirpc even on older versions of glibc - it is capable of handling IPv6.
    Use pkg-config to find out the correct include path and library to add.
    

provision/initramfs/Makefile.am
```diff
@@ -53,7 +53,10 @@ busybox:
 		for i in $(BUSYBOX_PATCHES); do \
 			(cd _work/$(BUSYBOX_DIR); patch -p1 < ../../$${i} || exit 1) \
 		done ;\
-		cp $(srcdir)/busybox.config _work/busybox-$(BUSYBOX_VERSION)/.config ;\
+		sed -e 's/"/\\"/g' \
```
bensallen: This `sed` command is a bit dense. Can you explain what it's doing?

bensallen:
Should we add a dependency into the RPM spec file for libtirpc-devel?

bensallen:
A higher level question, what Busybox functionality are we using today or are you needing to use that requires RPC support? If we're not actually using it can we just disable applicable Busybox config option(s)?

jjolly:
> A higher level question, what Busybox functionality are we using today or are you needing to use that requires RPC support? If we're not actually using it can we just disable applicable Busybox config option(s)?

Disbling `CONFIG_FEATURE_HAVE_RPC` removes the inet and mount features from busybox, and I'm pretty sure mount, at a minimum, is necessary.

So, considering the need for libtirpc, I will likely make a change to the patch to put the pkg-config work into the configure.ac. Please hold on this PR until I've made that change.

jjolly:
Changed things around to get pkg-config data in configure and stuff it into the busybox.config. This required renaming `busybox.config` to `busybox.config.in`

Please review.

bensallen:
You'll need to add libtirpc-devel to https://github.com/warewulf/warewulf3/blob/master/.circleci/config.yml#L16

PR #131, opened 2018-09-11
Suse prov apache2 conf use suse sysconfdir

jjolly:
Check for SUSE system and set Apache2 config path accordingly
    

provision/etc/Makefile.am
```diff
@@ -7,9 +7,13 @@ bash_completiondir = $(sysconfdir)/bash_completion.d/
 if ISDEB
 httpdir = $(sysconfdir)/apache2/conf.d/
 else
+if ISSUSE
```
bensallen: I ❤️ Automake syntax for if else.

PR #137, opened 2018-09-11
Suse prov dhcpd add vendor class identifier

jjolly:
Add DHPC vendor-class-identifier option for iPXE
    
    With multiple DHCP server present, iPXE may chose the wrong one
    if no option vendor-class-identifier is present in DHCP offer.
    
### DIFF
--- a/provision/etc/dhcpd-template.conf
+++ b/provision/etc/dhcpd-template.conf
@@ -12,6 +12,9 @@ option ipxe.no-pxedhcp 1;
 option architecture-type   code 93  = unsigned integer 16;
 
 if exists user-class and option user-class = "iPXE" {
+    if exists vendor-class-identifier {
+         option vendor-class-identifier = option vendor-class-identifier;
+    }
     filename "http://%{IPADDR}/WW/ipxe/cfg/${mac}";
 } else {
     if option architecture-type = 00:0B {


bensallen:
@jjolly Can you explain a bit more here whats going on? Is this setting the vendor class identifier field of the server's DHCP offer to the same as what was in the client's DHCP request?

Any reference from iPXE showing that it would preferentially pick one offer over another due to vendor class identifier field being populated?

e4t:
This was an observation I've made experimentally. I've got 2 DHCP servers running - one is serving addresses for 'normal' boots, the 2nd one for PXE boots. The first one cannot be configured, thus it will reply to any dhcp request. The 2nd one is the ISC dhcp server. Most PXE boots ignore the the first one as it doesn't provide a boot file. The iPXE one doesn't: it stubbornly picks the primary one that doesn't provide a boot file and then is unable to proceed. When enabling this option the ISC DHCP server sends back the vendor class identifier it has received: when this is the case, the iPXE boot will pick this offer.
I did run into issues, however, with other PXE implementations sending the vendor class identifier back with the offer - this is why I like to limit this to the iPXE implementation,
Admittedly I've  only debugged this using a network sniffer I didn't check this behavior in the iPXE code.

bensallen:
Related code in iPXE is at https://github.com/ipxe/ipxe/blob/master/src/net/udp/dhcp.c#L385 and below. Possibly here as well: https://github.com/ipxe/ipxe/blob/master/src/net/udp/dhcp.c#L584.

I *think* by setting vendor class id of the DHCP offer to PXEClient... you're tricking iPXE into thinking it's dealing with ProxyDHCP (eg. has_pxelclient is set) in dhcp_discovery_rx allowing it to transition state from dhcp_discovery_rx to dhcp_request_tx. Note, iPXE has the request's vendor class id here: https://github.com/ipxe/ipxe/blob/master/src/include/ipxe/dhcp.h#L217, which is indeed starts with PXEClient.

Since the passed dhcp_session in dhcp_request_tx has the proxy_offer attributed populated, it doesn't go on to attempt a ProxyDHCPREQUEST.

Let me know if your take matches above.

If you have access to the environment where you can reproduce this behavior, I would try adding to dhcpd.conf. This appears as it will have the effect as tricking iPXE into thinking its dealing with proxy DHCP:

```
# Tell iPXE to not wait for ProxyDHCP requests to speed up boot.
option ipxe.no-pxedhcp code 176 = unsigned integer 8;
option ipxe.no-pxedhcp 1;
```

Otherwise, building ipxe with make ... DEBUG=dhcp would give us a better sense of whats going on.

e4t:
After digging into this more deeply:
The PXE spec http://download.intel.com/design/archives/wfm/downloads/pxespec.pdf) describes a 2-step process to receive the client IP address and the boot file. The client IP can be obtained from a 'standard' DHPC server (without knowledge of PXE) or a PXE-aware one. If a standard server is to be used, the PXE info (about boot servers etc.) must be sent from a proxy-DHCP server. 
The boot file is obtained by a 2nd discovery (p.14 step 5). What happens if the initial offer contains a boot file already isn't covered - thus Warewulf (and many others) use PXE out of spec. This is where the Intel implementation (used by most DHCP clients) and iPXE differs:
the Intel implementation will get preference to those offers which contain a boot file, the iPXE implementation waits until it times out and picks the offer that happened to come in last - of course, I'm not sure what the Intel implementation really does under the hood.
Changing ipxe to give an offer precedence which has a either a boot file or a boot menu set fixes this.
Without touching ipxe, I'm afraid the only way to fix this is to make it think if is talking to a proxy.
I will contact the ipxe guys to see what they think about changing this - since it will affect something that has been left random should not harm anything.
Another way of doing this - which is even better if ipxe cannot be changed - would be to user Etherboot priorities:
<pre>
option space gpxe;
option gpxe-encap-opts code 175 = encapsulate gpxe;
option gpxe.priority code 1 = signed integer 8;
host ... {
  ...
  option gpxe.priority 2;
}
</pre>

bensallen:
iPXE has the equivalent setting: ` option ipxe.priority code 1 = signed integer 8;` Ref: https://ipxe.org/howto/dhcpd#ipxe-specific_options. Although I've not experimented with this setting.

If you look at the commit history of src/net/udp/dhcp.c it appears there were attempts at what you suggest, but they were reverted long ago.

bensallen:
Closing this PR as I don't think the workaround included is something we want to include in Warewulf.

PR #140, opened 2018-09-11
Suse prov config local binary copy

jjolly:
provision: allow build to use locally-installed busybox
    
    Traditionally, distribution vendors have their own build of the Busybox
    utility. Having Warewulf build a specialized copy of Busybox causes
    issues:
    * Duplicate builds cause unnecessary extra time and resources to be
      spent in the build process
    * Patches need to be applied and tested twice
    
    To mitigate this issue, two extra configuration items have been added
    to the provision configure script:
    * --with-local-busybox allows the builder to specify the location of
      the busybox binary
      - If no path to the binary is specified, then the PATH is searched
        for a suitable binary
      - If a suitable binary cannot be found, the system falls back to
        building Busybox internally
    * --with-busybox-links-file specifies the list of links that need to be
      created along with the Busybox binary
      - If a local Busybox binary is found, this defaults to
        /usr/share/busybox/busybox.links
      - This file must exist for the local Busybox binary to be used
    
    In all cases, if these configuration items are not used, the build
    falls back to building Busybox internally.
    
    Signed-off-by: John Jolly <jjolly@suse.com>

### DIFF
--- a/provision/3rd_party/Makefile.am
+++ b/provision/3rd_party/Makefile.am
@@ -21,19 +21,37 @@ prep:
 bin-i386-pcbios/undionly.kpxe: prep
 
 if BUILD_X86_64
-	$(MAKE) -C _work/$(IPXE_DIR)/src CROSS_COMPILE=$(CROSS_COMPILE_X86_64) bin-i386-pcbios/undionly.kpxe
+	@ if [ -n "@local_ipxe_undionly_path@" -a -f "@local_ipxe_undionly_path@" ]; then \
+		echo "Detected local install of undionly.kpxe boot image. Bypassing build process." ;\
+		mkdir -p _work/$(IPXE_DIR)/src/bin-i386-pcbios ;\
+		cp "@local_ipxe_undionly_path@" _work/$(IPXE_DIR)/src/bin-i386-pcbios/undionly.kpxe ;\
+	else \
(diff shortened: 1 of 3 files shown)


bensallen:
We also can build the arm64 build of iPXE. Not sure if any distro packages it at this point, but it might make sense to add finding it locally on the system as well.

provision/3rd_party/Makefile.am
```diff
@@ -12,28 +12,52 @@ IPXE_DIR = ipxe-$(IPXE_VERSION)
 
 
 prep:
-	@ if [ ! -d "_work/$(IPXE_DIR)" ]; then \
-		echo "Extracting IPXE" ;\
-		mkdir -p _work/ ;\
-		tar -xJf $(IPXE_SOURCE) -C _work/ ;\
+	@ if [ -n "@local_ipxe_snp_i386_path@" -a -f "@local_ipxe_snp_i386_path@" -a \
```
bensallen: Would it be reasonable to have a single autoconfig variable that signifies local_ipxe or not local_ipxe instead of a variable for each binary, so we don't have so much code in the Makefile for this?

Also, I'm thinking if local_<app> is defined and the file doesn't exist we should bail out as an error instead of falling back to building the internal version. Otherwise we're going to be hiding a potential mismatch between what the configure script found and the current state of the system.

jjolly:
I took out most of the conditions from the Makefile.am, and now I'm just copying the file as needed into the final build location. This allows the original loop to copy the files from the build area, without regard to if it was built locally or came from a local install.

Looks cleaner to me. Let me know what you think.

PR #141, opened 2018-09-11
Suse prov config local haveged copy

jjolly:
provision: Install HAVEGEd to provide entropy to parted during bootstrap
    
    When creating a stateful system during the bootstrap process, the
    parted program is used to partition the local storage device. The
    parted program uses data from the kernel pseudorandom number generator
    to create unique identifiers. On certain systems, such as in KVM
    Virtual Machines, not enough entropy is provided for the parted
    program. This causes the parted program to hang indefinitely waiting
    for entropy to accumulate.
    
    The HAVEGEd daemon provides addition entropy based on system actions.
    If started early enough in the bootstrap process, the HAVEGEd daemon
    provides sufficent entropy so that the parted program will finish
    successfully.
    
    This fix searches for the haveged program and, if found, adds haveged
    to the initramfs.cpio bootstrap file. This search only occurs if the
    --with-local-haveged option is used during build configuration. If this
    option is not provided, or the haveged program cannot be found on the
    system, then haveged is not copied to the initramfs.cpio bootstrap file
    and no attempt is made to execute haveged during the bootstrap init.
    
    Signed-off-by: John Jolly <jjolly@suse.com>


jjolly:
As discussed, I plan to receive approval for HAVEGEd to be included to the warewulf source packages. For now, though, I'd like to keep this selectable feature as it will be useful for distributions after HAVEGEd has been added.

PR #142, opened 2018-09-11
Suse vnfs zypper auto agree with licenses

jjolly:
vnfs: Add --auto-agree-with-licenses to include-suse PKGR_CMD
    
    When attempting to create an opensuse chroot, the wwmkchroot script
    would fail on packages that attempt to display a license and expect a
    response from the console.
    
    The zypper package manager has an option that automatically accepts
    package license agreements when present. Adding this option to the
    PKGR_CMD variable resolved the issue with mkchroot.
    
    Signed-of-by: John L. Jolly <jjolly@suse.com>


jjolly:
> Does this still display the license to the user or does it hide it?

The `--auto-agree-with-licenses` option does not display the license, but rather skips the entire license display-and-agree process. The assumption is, when using the option, you know what you are doing and are, by default, agreeing to the licenses in the package.

This may not help in the case of wwmkchroot. Would it be better if the zypper tool had the ability to display the license and auto-accept?

jjolly:
For completeness, when the `--auto-agree-with-licenses` option is used, the one line of text related to licensing that is displayed is:

    Automatically agreeing with openSUSE Tumbleweed product license.

bensallen:
> Would it be better if the zypper tool had the ability to display the license and auto-accept?

In other distros we don't have a license agreement acceptance functionality in the package utility, so we're not taking the extra step of implicitly agreeing license agreements in place of the user. Here we are implicitly agreeing for the users.

So for our use-case, yes this functionality would be useful. This approach will cover the Warewulf project a bit better, as we're making a good faith effort to show the users the license.

In the interim though, I think `Automatically agreeing with openSUSE Tumbleweed product license.` is good enough.

Let me know if the zypper functionality you're suggesting is easy to add in the timeframe of this PR. If its longer term, I'll approve this PR as-is.

Ben

jjolly:
> Let me know if the zypper functionality you're suggesting is easy to add in the timeframe of this PR. If its longer term, I'll approve this PR as-is.

The chance is highly unlikely that we will get the feature into zypper very quickly. If it were for me to decide (I'm not the zypper maintainer, so not likely), I would add another option to zypper such as `--always-display-license` which would display the license even if auto-accept was enabled.

For now, if you could accept the change as it stands, we would appreciate that. When (if) zypper gets changed, we'll submit another PR using that change.

bensallen:
Sounds good. Can you resolve the conflict on include-suse?

PR #144, opened 2018-09-15
common: Correctly detect SUSE system for system services

e4t:
Since /etc/SuSE-release was removed the detection of SUSE systems for systemd services didn't
work correctly any more. It was replaced by /etc/os-release.

### DIFF
--- a/common/lib/Warewulf/SystemFactory.pm
+++ b/common/lib/Warewulf/SystemFactory.pm
@@ -47,13 +47,24 @@ new($$)
     my ($mod_name, $obj);
 
     if (! $type) {
+        $type = "unsupported";
         if (-f "/etc/redhat-release") {
             $type = "rhel";
         } elsif (-f "/etc/SuSE-release") {
             $type = "Suse";
-        }
-		else {
+        } elsif ( -f "/etc/debian_version" ) {
             $type = "Deb";
+        } elsif ( -f "/etc/os-release") {
+            open(SYSTEM, "/etc/os-release");
+            while (<SYSTEM>) {
+                if (/^NAME=/) {
+                    if (/SUSE|SLE/) {
+                        $type = "Suse";
+                    }
+                    last;
+                }
+            }
+            close SYSTEM;
         }
     }
 


common/lib/Warewulf/SystemFactory.pm
```diff
@@ -51,8 +51,21 @@ new($$)
             $type = "rhel";
         } elsif (-f "/etc/SuSE-release") {
             $type = "Suse";
+        } elsif ( -f "/etc/os-release") {
+            open(SYSTEM, "/etc/os-release\n");
```
bensallen: Extra newline in the file path?

common/lib/Warewulf/SystemFactory.pm
```diff
@@ -51,8 +51,21 @@ new($$)
...
+            while (<SYSTEM>) {
+                if ($_ =~ /NAME=/) {
```
bensallen: Perhaps `/^NAME=/` here?

common/lib/Warewulf/SystemFactory.pm
```diff
@@ -51,8 +51,21 @@ new($$)
...
+                    if ($_ =~ /SUSE|SLE/) {
+                        $type = "Suse";
+                    }
+                    last;
+                }
+            }
+            close SYSTEM;
+            goto NOSYS if (! $type);
```
bensallen: If you could help clean up our existing code a bit, instead of using a `goto` to get into the catch all `else`, add a conditional before `} elsif ( -f "/etc/os-release") {` to check for `/etc/debian_version` and set $type = "Deb". Then under the catch all `else` set type to `unsupported`.

common/lib/Warewulf/SystemFactory.pm
e4t: Fixed.

common/lib/Warewulf/SystemFactory.pm
e4t: Fixed now.

common/lib/Warewulf/SystemFactory.pm
e4t: Fixed now.

PR #161, opened 2018-11-08
Remove bogus "* checksum ..." line from wwgetvnfs

novosirj:
Was generating an error on a booted node in /var/log/warewulf/provision/getvnfs.log -- looks to have been added by mistake. Mostly harmless, but still wrong if I'm reading this right.


novosirj:
Accidentally did this incorrectly.

PR #173, opened 2019-01-29
Manually configure master IP/network/netmask

beinvisible:
Allow direct configuration of IP/network/netmask for DHCP/PXE in provision.conf.

This is useful for systems where we have no control over the naming of network interfaces and a DHCP server running on the wrong interface might have unpleasant side effects.

Settings are completely optional; if not set, addresses are derived from 'network device' setting as before.

### DIFF
--- a/provision/etc/provision.conf
+++ b/provision/etc/provision.conf
@@ -3,6 +3,13 @@
 # communicate with the nodes?
 network device = eth1
 
+# Alternatively configure the network addresses directly, e.g. if the
+# network device name might change. This overrides the values derived 
+# from the 'network device' setting (if configured) one by one.
+# ip address = 172.16.1.1
+# ip netmask = 255.255.0.0
+# ip network = 172.16.0.0 
+
 # Which DHCP server implementation should be used?
 dhcp server = isc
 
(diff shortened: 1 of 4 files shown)


beinvisible:
I haven't changed anything about the old behavior, so a missing 'network device' configuration would result in just the same errors as before my patch if there is no ip address/network/netmask configured. And those options are not mutually exclusive to the 'network device' setting - missing values will still be derived from there, so we might end up with a mix. Depending on your point of view this can be a bug or a feature (e.g. you might only want to serve a part of the network device's subnet and therefore just configure a different netmask).

As far as I can see, the network device itself is never used; it's always the derived address.

For example the Isc.pm modeule checks if all 3 items (address, network, netmask) are set. Also the dhcpd.conf template only uses the addresses but not the network device itself.

bensallen:
@beinvisible 

> I haven't changed anything about the old behavior, so a missing 'network device' configuration would result in just the same errors as before my patch if there is no ip address/network/netmask configured.

Yep, just wondering if you had tested those three conditions with your change to double check.

> As far as I can see, the network device itself is never used; it's always the derived address.

Agreed, I don't believe it's being used anywhere else.

Could you add commented out examples into provision/etc/provision.conf with these new settings, so users know they exist? Ideally with a little bit of text stating "Overrides IP Address derived from network device" or similar?

beinvisible:
Tested various combinations and it works as expected.

Also added the comment in provision.conf and updated the error messages.

PR #179, opened 2019-03-15
Fix: renamed pxelinux attribute

beinvisible:
pxelinux was renamed to pxeloader

The Dnsmasq module still used the old method - a copy&paste artifact from Isc module in master. Dnsmasq does not support overriding the pxe boot image per node, so I removed the unnecessary call.

Also the wwsh manpage did not reflect the changed parameter name.


bensallen (approved):
Thanks.

PR #195, opened 2019-07-02
adding ipxe chainloading from VirtualBox netboot that is also using ipxe

truatpasteurdotfr:
- netbooting with virtualbox is using ipxe and send the iPXE tag which is caught by the dhcpd-template.conf
- but the vbox provided version ipxe version  is size limited and lack the capacity to boot from  initrd/kernel provided by warewulf setup
- I propose to change de iPXE tag with a "warewulf" tag and chainload the "warewulf" tagged undionly.kpxe


bensallen:
We’ll need to dig into this further.

I don’t want us to change the default boot behavior to work around VirtualBox’s iPXE. Especially when the Intel PXE bootrom from the extension pack works as expected.

Is there any other way to detect when iPXE doesn’t support bzImage, eg if I recall the issue with VirtualBox’s iPXE? Either with DHCP options or within an iPXE script?

bensallen:
Looks likes there are DHCP options for iPXE features: http://ipxe.org/howto/dhcpd#ipxe-specific_options.

So we could add a conditional on: if iPXE and ( bzImage or EFI iPXE feature support ) else chainload appropriate iPXE binary.

truatpasteurdotfr:
yes, that would also be ok!

truatpasteurdotfr:
less intrusive solution provided

Issue #104, opened 2018-01-30
Mellanox FlexBoot does not send user-class "iPXE" in DHCP request


bensallen:
We're using the DHCP user-class option 76 from a DHCP request to identify clients as iPXE ( (https://github.com/warewulf/warewulf3/blob/master/provision/etc/dhcpd-template.conf#L14). 

Mellanox FlexBoot is a modified iPXE, so we expected it would identify itself via this option as well. This doesn't appear to be case. An alternative method needs to identified to conditionally setting the filename option to the iPXE cfg file for FlexBoot.

Discussion and debugging in #100.

truatpasteurdotfr:
I have a similar issue, with virtualbox (without the extension pack) pxe booting with a stripped down ipxe version which trigger the matching iPXE tag, but without the bzImage loading capacity...

PR #196, opened 2019-07-02
Detect iPXE features

bensallen:
Detect if iPXE has the features needed for Warewulf:
- ipxe.http - fetch the kernel and initrd over http
- ipxe.bzimage - Linux bzImage format kernels
or
- ipxe.efi - Linux kernel supporting EFI_STUB, eg. for EFI boot support

If features are missing (including if its not iPXE at all) chainload our iPXE based on architecture.

### DIFF
--- a/provision/etc/dhcpd-template.conf
+++ b/provision/etc/dhcpd-template.conf
@@ -3,15 +3,23 @@ allow bootp;
 ddns-update-style interim;
 authoritative;
 
+# Declare the iPXE option space
 option space ipxe;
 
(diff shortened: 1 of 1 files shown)


bensallen:
@truatpasteurdotfr please give this a test.

bensallen:
I suspect this will actually close #104, but I don't have an environment handy to test with right now.

truatpasteurdotfr:
that works for me :P

truatpasteurdotfr (approved):

PR #200, opened 2019-09-05
Fixed stateful provisioning UEFI boot settings for aarch64

NaohiroTamura:
@bensallen thanks for the review.


NaohiroTamura:
Closed due to wrong operation, it intended to update #199.

PR #201, opened 2019-09-30
Centos8 Updates

jcsiadal:
Initial updates for CentOS 8. I built a head node and one compute node with this update. Normally, I would do more testing and also recheck against the other distros, but I thought it would be helpful to make this available while I test on CentOS 7, Ubuntu 18, and SLES 15.

Summary of changes

- Updated WW version in configure.ac to expected release version.

- ntp no longer exists in RHEL. I normally test "wwinit all" and the ntp module was throwing an error and then configuring a default ntp.conf file. The module was updated to ignore ntp if it sees chrony installed. TODO: Create a chrony wwinit module; I've been manually configuring it.

- I originally added logic to internally branch the include-redhat script, but it makes it more difficult to edit and read. Instead, I've added a new includes for dnf-based redhat. One major change is a requirement to add a second repo for installation. Without a second repo, the admin would need to manually merge the AppStream and BaseOS repos from the ISO or web.

- Update to wwvnfs to dynamically check for sysmacros.ph before loading. The sysmacros.h file no longer exists in glibc 2.26, so Red Hat removed the .ph from Perl. I'm not even sure its needed on previous distros. I looked at changing the configure and make scripts to update wwvnfs based on OS, but the dynamic check was more expedient.

- Updated scripts in the initrd to dump errors to a log file instead of null. I hit some problems during the provisioning process and its helpful to have error output for review.

- Added my frontend build script to /platform. Again, this is not intended for inclusion in the release, but to provide help identify WW dependencies and build options.

### DIFF
--- a/cluster/configure.ac
+++ b/cluster/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ(2.59)
-AC_INIT(warewulf-cluster, 3.8.2, warewulf-devel@lbl.gov)
+AC_INIT(warewulf-cluster, 3.9, warewulf-devel@lbl.gov)
 AC_CONFIG_SRCDIR([.])
 
 AC_PROG_INSTALL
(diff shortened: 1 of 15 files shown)


jcsiadal:
The build check failed. I'm closing this PR and will resubmit. I want to split the CentOS 8-specific updates from the more general updates. I also had a last-minute change-of-mind; I want to align include-redhat with the multi-repo method used in include-suse.

PR #202, opened 2019-10-02
Add error capture to initrd HTTP transport scripts

jcsiadal:
If a wget request fails when booting the node, there is a 30s window to manually login and investigate. The problem is that the http-transport scripts output errors to null.

This update redirects the output to /tmp/http.log instead.

### DIFF
--- a/provision/initramfs/capabilities/transport-http/wwgetfiles
+++ b/provision/initramfs/capabilities/transport-http/wwgetfiles
@@ -54,9 +54,10 @@ elif [ -f "$TIMESTAMP_FILE" ]; then
 fi
 
 while true; do
+    echo "== wwgetfiles ==" >> /tmp/http.log
     for master in `echo $WWMASTER | sed -e 's/,/ /g'`; do
         TMPFILE=`mktemp`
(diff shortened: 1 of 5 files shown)


provision/initramfs/capabilities/transport-http/wwgetfiles
```diff
@@ -54,9 +54,10 @@ elif [ -f "$TIMESTAMP_FILE" ]; then
 fi
 
 while true; do
+    echo "== wwgetfiles ==" >> /tmp/http.log
```
jmstover: It may be good to set this prior to the `while true; do` ... IIRC you would end up with multiple `== wwgetfiles ==` entries in the `http.log` file if the connection to the master timed out.

jmstover (dismissed):
Beyond my one comment LGTM

@bensallen -- See any issues?

provision/initramfs/capabilities/transport-http/wwgetfiles
jcsiadal: You're right. Let me adjust that and force a push.

jmstover (dismissed):

bensallen:
@jmstover Don't we log to /var/log/... typically in the initramfs?

jcsiadal:
I'm going to rebase this and change the log file location to /var/log/warewulf/provision/http.log.

bensallen (approved):

PR #203, opened 2019-10-03
Add CentOS 8 Support

jcsiadal:
Adds support for CentOS 8 (should also support RHEL 8, but not yet tested). There is a change in chroot creation that needs to be reviewed and approved.

List of changes and details
Update WW versions in autoconfigure to 3.9

- Update the wwinit NTP script to terminate if chrony is found. There is no "ntp" RPM in CentOS 8, so was producing an error. TODO: Add git issue to create a 51-chrony.init module.
- Include my platform build script for CentOS 8. Not intended for release, but for development. It's important to note that I needed to use more OS-provided components, as the included sources would not compile in CentOS 8. I think a good future project is to detect the target OS and do this during autoconfigure.
- Update the wwvnfs script to check for sysmacros.ph. Glibc 2.26 drops sysmacros.h, so the Perl conversion goes away in CentOS 8. I looked at different ways to handle this, since I don't know if its needed in any modern distro. My quick fix is to try and load it, and then actually load it if no error occurs. This avoids alternate scripts for different Linux versions.
- Add wwmkchroot support for CentOS 8.

This last item is a change in how the vnfs chroots are built. RHEL 8 moved to DNF, which _mostly_ supports YUM options. The current DNF really impacted operation in chroots.

The big change here was I decoupled the chroot package manager configuration from the host pkgmgr configuration. WW 3.8 copies the head node YUM config/repos to the chroot before installation, using a temporary YUM config file. I removed that. Instead, it makes the repos and configuration used during chroot installation permanent within that chroot. The head node and compute node repos remain separate. This allows me to easily use a CentOS 8 or RHEL 8 head node and build CentOS 6, 7, or 8 chroot images. I tested building all three of these.

I'll need to do some more work to make sure these cross-built chroots are maintainable on their own -- that is, updating packages in the chroot after creating it. I'd also like to confirm cross-building from SUSE or Ubuntu.

### DIFF
--- a/cluster/libexec/wwinit/50-ntpd.init
+++ b/cluster/libexec/wwinit/50-ntpd.init
@@ -20,9 +20,14 @@ fi
 
 wwreqroot
 
+if  rpm -q chrony >/dev/null 2>&1 ; then
+    wwprint "Chrony found; skipping NTP configuration\n"
+    exit 0
+fi
+
 if [ ! -f "/etc/ntp.conf" ]; then
     wwprint "Is NTP installed? /etc/ntp.conf is not present!\n" error
-    return 1
+    exit 1
 fi
 
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,6 @@ jobs:
           destination: SRPMS
 
   build_centos8:
-
     docker:
      - image: centos:8
 
@@ -159,10 +158,4 @@ workflows:
   build:
     jobs:
       - build_centos7
-         - filters:
-            tags:
-              ignore: el8
       - build_centos8
-         - filters:
-            tags:
-              only: el8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,8 +76,93 @@ jobs:
           path: ~/rpmbuild/SRPMS
           destination: SRPMS
 
+  build_centos8:
+
+    docker:
+     - image: centos:8
+
+    working_directory: ~/warewulf3
+
+    steps:
+      - checkout
+      - run:
+          name: Install Build Dependencies
+          command: |
+            set -o xtrace
+            dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+            rpmkeys --import /etc/pki/rpm-gpg/RPM-GPG-KEY-*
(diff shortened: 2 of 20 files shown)


bensallen (changes requested):
Please change the version to 3.9.0 instead of 3.9. Also the CircleCI config needs updating to get build to pass: https://github.com/warewulf/warewulf3/blob/master/.circleci/config.yml#L34. Certainly not required for this PR, but if you are so inclined it would be amazing to also have a build_centos8 job in CircleCI.

cluster/libexec/wwinit/50-ntpd.init
```diff
@@ -20,9 +20,14 @@ fi
 
 wwreqroot
 
+if  rpm -q chrony >/dev/null 2>&1 ; then
```
bensallen: @jmstover Thoughts on if we should bail from 50-ntpd.init if chrony is installed? It is the default in EL7 and 8, but at least in my environment we still use ntpd everywhere. Alternatively we could reverse the order, and check for ntp.conf, if not found and chrony *is* found, exit 0 cleanly. If ntp.conf isn't found *and* chrony is not found then we exit 1.

jcsiadal: Sounds fine to me. Admins shouldn't have both installed at the same time, but I just confirmed you can. Can you confirm these changes will do what you're asking? Changing the filenames so there's a clear order to wwinit.

1. 51-chrony.init: if ntp or openntpd found then return error 0.
2. 51-chrony.init: if chrony not found, return error 1.
3. 52-ntp.init: if ntp/openntpd not found, check for chrony.
4. 52-ntp.init: if chrony found, return error 0, else return error 1.

jcsiadal:
I updated the version # and ntpd init. In addition, I added a workaround for new Dracut to the mkchroot, if a kernel is included at build-time. Unfortunately, it doesn't help with kernel installs on a working CentOS8 chroot.

bensallen:
> Please change the version to 3.9.0 instead of 3.9.

@jcsiadal Looks like the version is now 3.90 instead of 3.9.0. Can you fix this to set the version to 3.9.0?

bensallen:
https://github.com/warewulf/warewulf3/blob/master/.circleci/config.yml#L34 also needs to be updated to install the current RPMs based on 3.9.0 to get the CircleCI builds passing.

jcsiadal:
Fixed version to 3.9.0.
Updated the rpm spec.in files as needed.
Updated the circleci config. No simple way to test it on my side, and my first attempt is broken. Since I have the error output, I'll work on getting this fixed.

jcsiadal:
Leaving the PR in place right now, because I need the CI tests running. I don't know how to make that happen outside of a PR. When this is working, I'll close the PR, rebase, squash the commits, and resubmit.

jcsiadal:
Replaced by #207

PR #208, opened 2019-10-21
proper locking in wwgetfiles script

beinvisible:
The wwupdatefiles cronjob leaves a lockfile behind if it is killed for whatever reason, preventing it from ever running again. Also the locking mechanism isn't really atomic.

This patch changes locking to use the 'flock' command.

### DIFF
--- a/provision/initramfs/capabilities/transport-http/wwgetfiles
+++ b/provision/initramfs/capabilities/transport-http/wwgetfiles
@@ -12,13 +12,9 @@ if [ -n "$SET_X" ]; then
 fi
 
 # Lock so more than one wwgetfiles cannot be ran
-LOCKF=/tmp/.wwgetfile.lock
-if [ ! -f ${LOCKF} ]; then
-    touch ${LOCKF}
-else
-    echo "ERROR: wwgetfiles already running"
-    exit 1
-fi
+exec 9>/tmp/.wwgetfile.lock
+flock -n 9 || { echo "ERROR: wwgetfiles already running" ; exit 1; }
+echo $$ 1>&9
 
 if [ -f /warewulf/transports/http/functions ]; then
     . /warewulf/transports/http/functions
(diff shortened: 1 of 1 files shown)


provision/initramfs/capabilities/transport-http/wwgetfiles
```diff
@@ -108,14 +104,10 @@ while true; do
                 else
                     echo "ERROR: Could not download $name"
                     rm -f $NEWROOT/$path.ww $NEWROOT/$path.ww$timestamp
-                    # Remove lock file
-                    rm -f ${LOCKF}
                     exit 2
```
jmstover, line 108: Add in a `exec 9>&-` prior to the exit here? Just to be explicit on releasing the lock ...

jmstover (changes requested):

beinvisible:
Why? On exit the open filehandles (and locks on them) are released anyways.

But thinking about it, I'd actually prefer the block/subshell-syntax 
suggested in `man flock`, but I'm not sure if all variants of /bin/sh 
would accept it.

provision/initramfs/capabilities/transport-http/wwgetfiles
beinvisible: Why? On exit the open filehandles (and locks on them) are released anyways.

I'd actually prefer the block/subshell-syntax suggested in `man flock`, but I'm not sure if all variants of /bin/sh would accept it.

jmstover: Call it paranoia mostly. It _should_ be released yes...

beinvisible: If a terminated process can hold filehandles and locks then something is very wrong with the system.

bensallen: How about `flock -u -n 9` and log explicitly if it fails. On the very odd instance that this happens, we'd at least have a log message to debug with then. Letting the script just exit means we loose the chance to log it.

beinvisible: >         -u, --unlock
>              Drop a lock.  *This is usually not required, since a lock is automatically  dropped when*
>              *the  file  is  closed.*  However, it may be required in special cases, for example if the
>              enclosed command group may have forked a background process which should not be  holding
>              the lock.

We do not fork background processes (and even if we did, they *should* hold the lock).

bensallen: Fair enough.

bensallen (approved):

jmstover (approved):

Issue #198, opened 2019-08-07
bootstrap filesystem failing cause /dev/sd* not ready


renatopancheri:
Hi,

i encountered on some machines a weird issue,
the filesystem script at boot fails with the error:

>  /usr/sbin/parted --align=optimal --script /dev/sda mklabel gpt mkpart primary linux-swap 0% 20% mkpart primary ext4 20% 100% name 1 swap name 2 scratch
Error: Could not stat device /dev/sda - No such file or directory.

the disk is actually present and functioning correctly once the machine has finished booting,
see this:
> [root@n0170 ~]# dmesg|grep -i sda
[    9.481470] sd 6:0:0:0: [sda] 1953525168 512-byte logical blocks: (1.00 TB/931 GiB)
[    9.481878] sd 6:0:0:0: [sda] Write Protect is off
[    9.481885] sd 6:0:0:0: [sda] Mode Sense: 00 3a 00 00
[    9.482094] sd 6:0:0:0: [sda] Write cache: enabled, read cache: enabled, doesn't support DPO or FUA
[    9.523815]  sda: sda1 sda2
[    9.524854] sd 6:0:0:0: [sda] Attached SCSI disk

it seems that the filesystem script is called before the kernel adds /dev/sda to the device tree.
The solution i've come up is to create and adhoc pre script to wait for /dev/sda to be ready.
Anyone else has a better idea on how to better handle this situation?

bensallen:
One possible workaround is adding whatever kernel module is being used the controller of sda to `wwkmods=<mod>` in kargs for the host. These are loaded slightly sooner in the boot process compared to when the detect script iterates the PCI bus looking for drivers to load.

Otherwise, we'll probably need to add a check for the device and add a retry mechanism when the 'select sda' is specified: https://github.com/warewulf/warewulf3/blob/development/provision/initramfs/capabilities/setup-filesystems/20-filesystems#L161.

renatopancheri:
Hi, I've finally found the time to look into this thing and i added a check in the 20-filesystems file:
`206,214d205`
`<       if [ "${SELECT_DISK##*/dev/*}" == "" ] && [ ! -b ${SELECT_DISK} ]; then`
`<         msg_gray "   * ${SELECT_DISK} not ready, waiting 5 seconds..."`
`<         sleep 5`
`<         wwsuccess`
`<         if [ ! -b ${SELECT_DISK} ]; then`
[...]

PR #214, opened 2020-01-03
Added check of /dev/sd* before running parted

renatopancheri:
Isssue #198 

Here is the pull request you asked, on a side note i use the prepared rpms from 
https://github.com/openhpc/ohpc
as far as i understand they use this repo tag 3.8.1 as sources as base and then apply some patches, it would be cool for me if every once in a while you merge into master and create a tag so that also the ohpc repo gets the patches from here more often, this way i can get the patches just by doing "yum upgrade", thanks in advise.


bensallen (approved):
Looks great. Thanks for the contribution.

Issue #216, opened 2020-01-31
cluster-env doesn't check permissions before trying to edit authorized_keys


solj:
My authorized_keys file is set to 400 and every time I try to login, cluster-env tries to write to the file unsuccessfully. I'm not sure what the *right* solution is but perhaps a suggestion to the user to add the appropriate lines would be better than just repeatedly failing.

$ ls -l /home/solj/.ssh/authorized_keys
-rw-------. 1 solj admin 192 Jan 31 13:19 /home/solj/.ssh/authorized_keys

/usr/bin/cluster-env: line 21: /home/solj/.ssh/authorized_keys: Permission denied

jmstover:
`authorized_keys` is set 600 in your output (i.e. you have write access)... is that a modified permission (probably from `cluster-env`)? 

What version of Warewulf are you running? And how did you install it (i.e. OpenHPC? Source?)?

jmstover:
PR #217 _should_ fix getting the error message. Maybe not how you're wanting though...

jmstover:
Hrmm... this should have been autoclosed when #217 was merged... Meh.

PR #217, opened 2020-02-01
authorized_keys permission check

jmstover:
Check that authorized_keys is writable before attempting to write into it.

Fixes: #216


cluster/bin/cluster-env
```diff
@@ -23,7 +23,13 @@ if [ ! -f "$HOME/.ssh/config" -a ! -f "$HOME/.ssh/cluster" ]; then
     echo "Configuring SSH for cluster access"
     install -d -m 700 $HOME/.ssh
     ssh-keygen -t rsa -f $HOME/.ssh/cluster -N '' -C "Warewulf Cluster key" > /dev/null 2>&1
+    if [ -f $HOME/.ssh/authorized_keys -a ! -w $HOME/.ssh/authorized_keys ]; then
```
bensallen: This is a tiny nit, but I generally prefer this style as I believe shellcheck will complain

```
if [ -f $HOME/.ssh/authorized_keys ] && [ ! -w $HOME/.ssh/authorized_keys ]; then
```

jmstover: Changed. May eventually go through with shellcheck at some point and clean it up the scripts to make it happy. If you wanted you could it could be added into CI to be sure we keep them consistent.

bensallen: Yea I've wanted to enforce Shellcheck for a while, but lack the cycles.

bensallen (approved):

cluster/bin/cluster-env
jmstover: I'll see if I can't shift some to get that started then here.

PR #219, opened 2020-02-11
Provision fix

jmstover:
The order of how the command was being looked for was incorrect. Make sure that we see if we're supposed to print out help text before looking to see if we have nodes.


bensallen (commented):
Do we still need the if (! $command ) check on line 259 to display help if we're explicitly forcing the command to be "help" if ARGV is empty?

jmstover:
> Do we still need the if (! $command ) check on line 259 to display help if we're explicitly forcing the command to be "help" if ARGV is empty?

Nope. That has been removed... I just scanned over it. :/

bensallen (approved):

PR #221, opened 2020-02-21
Fix srv/state directory in specfiles

jcsiadal:
Updates the specfiles to correctly implement using a different service file directory (still defaults to /var). This is required for OpenHPC.

Cleaned up tags to put them in the same order and added missing URL tags.

Removed the Provides for ipmitool, since the tool is installed in a WW subdirectory. Provides for e2fsprogs, parted, are libarchive are still included, as those tools are installed in regular OS locations.

Fixed the Description wording, as requested in a previous PR.

### DIFF
--- a/cluster/warewulf-cluster.spec.in
+++ b/cluster/warewulf-cluster.spec.in
@@ -1,5 +1,6 @@
 %{!?_rel:%{expand:%%global _rel 0.%(test "@GITVERSION@" != "0000" && echo "@GITVERSION@" || git show -s --pretty=format:%h || echo 0000)}}
 %define debug_package %{nil}
+%{!?wwsrvdir:%{expand:%%define wwsrvdir %{_localstatedir}}
 
 Summary: Warewulf - HPC cluster configuration
 Name: warewulf-cluster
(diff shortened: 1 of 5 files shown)


jmstover:
Is this going to do what you want when `wwconfig STATEDIR` gives you a different value than what you're trying to set `--localstatedir` to? I mean that _shouldn't_ happen in a consistent build... 

In a lot of the files after common, we don't even care about what `--localstatedir` is... but instead what `wwconfig -s STATEDIR` returns.

jcsiadal:
The WAREWULF_STATEDIR environment variable when warewulf-common is configured appears to control the output of wwconfig.

My reason for setting is everywhere: I fear that there's some corner-case I'm not seeing, so I added localstatedir configuration across all RPM specs, to keep all relevant variables in sync. I figure that if there's no corner case, it's just redundant and won't hurt anything.

Let me know if you would like me to pull the extra settings.

jmstover:
> The WAREWULF_STATEDIR environment variable when warewulf-common is configured appears to control the output of wwconfig.
> 

See for example `provision/configure.ac` ...  line 26'ish is the start. Then in `provision/initramfs/Makefile.am` you have lines like:

```
 install -d -m 755 $(DESTDIR)/$(WAREWULF_STATEDIR)/warewulf/initramfs/
```

> Let me know if you would like me to pull the extra settings.

No ... no need to pull. It just may not do what you're expecting since the control in every package except warewulf-common is from `wwconfig` output. So it's *possible* to have an incorrect location if you have a throwaway warewulf-common installed that isn't built as how you're wanting the packages to be.

It's probably unlikely ... but it's still within the realm of possibility. ;)

jmstover (approved):

PR #223, opened 2020-02-28
updated capabilities so that you can have fstab in provision files

renatopancheri:
Hello,

i had to provision some nodes with different fstab so i added fstab as a provision file,
i encountered the problem that the entry generated by provision/fs (like swap) are only added in the fstab in the chroot and i loose them if i provision the file,
my solution was to:

1. modify `50-config` so that it saves the generated entries in /etc/fstab.ww.generated_entries
2. modify `wwgetfiles` so that if the file path is /etc/fstab and i find in the file the line `#GENERATED_ENTRIES#` i put below this line the generated entries taken from /etc/fstab.ww.generated_entries

This worked fine for me, i used the variable `${WWNOFSTABUPDATE}` because it was already in place before.

Waiting for feedback.

### DIFF
--- a/provision/initramfs/capabilities/provision-vnfs/50-config
+++ b/provision/initramfs/capabilities/provision-vnfs/50-config
@@ -39,5 +39,6 @@ if [ -z "${WWNOFSTABUPDATE}" ]; then
     else
         cp -f /tmp/fstab "${NEWROOT}/etc/fstab"
     fi
+    cp /tmp/fstab "${NEWROOT}/etc/fstab.ww.generated_entries"
 fi
 
--- a/provision/initramfs/capabilities/transport-http/wwgetfiles
+++ b/provision/initramfs/capabilities/transport-http/wwgetfiles
@@ -97,6 +97,9 @@ while true; do
                         else
                             chmod $mode "$NEWROOT/$path.ww$timestamp"
                         fi
+                        if [ -z "${WWNOFSTABUPDATE}" -a "$path" = "/etc/fstab" -a -f "$NEWROOT/etc/fstab.ww.generated_entries" ]; then
+                            sed -i "/#GENERATED_ENTRIES#/r $NEWROOT/etc/fstab.ww.generated_entries" "$NEWROOT/$path.ww$timestamp"
+                        fi
                         mv "$NEWROOT/$path.ww$timestamp" "$NEWROOT/$path"
                         chown $uid "$NEWROOT/$path"
                         chgrp $gid "$NEWROOT/$path"


bensallen:
@renatopancheri curious what fstab entries are you creating that can not be done via `wwsh provision --filesystem ...`?

renatopancheri:
nfs mounts + panfs mount (parallel file system by panasas) here is an example of the entry:
`panfs://10.152.0.100/sw   /sw  panfs   rw,noauto,panauto,rmlist=(10.152.0.100),readahead-level=10,statahead-level=4,iscsi-sock-rsize=40960,iscsi-sock-wsize=40960,callback-network-allow=10.150.0.0/16      0      0`
I dont want to use systemd .mount files (i dont even know if i can do it with this entry specifically...) so i used fstab.

bensallen:
The following might work, create or add to a new filesystems config eg. /etc/warewulf/filesystem/panfs.wwfs:

```
# Tmpfs root (equivalent to 20-filesystems' default settings)
select tmpfs
fstab "" / tmpfs rw:relatime:seclabel:mode=555 0 0

select panfs://10.152.0.100/sw
fstab "" /sw panfs rw,noauto,panauto,rmlist=(10.152.0.100),readahead-level=10,statahead-level=4,iscsi-sock-rsize=40960,iscsi-sock-wsize=40960,callback-network-allow=10.150.0.0/16 0 0
```

Then `wwsh provision set --filesystem=/etc/warewulf/filesystem/panfs.wwfs <node>`

I think one issue with this is 20-filesystems is going to try to mount this in the initramfs: https://github.com/warewulf/warewulf3/blob/development/provision/initramfs/capabilities/setup-filesystems/20-filesystems#L130. 

Assuming that's not desired, we should probably modify 20-filesystems to look for the noauto option and skip mounting. 

Alternatively, what I do for NFS is bake the NFS fstab entries into the VNFS:

```
#GENERATED_ENTRIES#
nfsServer:/home /home nfs vers=3,hard,tcp,intr,timeo=20,retrans=10 0 0
```

renatopancheri:
second option is what i used to do but now i need to have different fstab so i cannot do that,
first option i think its cool, i do not want to mount in the `20-filesystem` stage so i might try to modify this pull request so that it checks for the noauto as you said

bensallen:
The other approach I've taken is with Warewulf files and POSTSCRIPTs:

```
# wwsh file print add_soft.sh
#### add_soft.sh ###################################################
add_soft_restricted.sh: ID               = 219
add_soft_restricted.sh: NAME             = add_soft.sh
add_soft_restricted.sh: PATH             = /root/add_soft.sh
add_soft_restricted.sh: ORIGIN           = /root/add_soft.sh
add_soft_restricted.sh: FORMAT           = shell
add_soft_restricted.sh: CHECKSUM         = 760b10fe1029581afdb28694b76914bc
add_soft_restricted.sh: INTERPRETER      = UNDEF
add_soft_restricted.sh: SIZE             = 229
add_soft_restricted.sh: MODE             = 0400
add_soft_restricted.sh: UID              = 0
add_soft_restricted.sh: GID              = 0

# wwsh file show add_soft.sh
#!/bin/sh
echo "Running adhoc post-scripts"
echo "Adding soft mount into fstab"
echo 'nfsServ:/vol/soft/x86_64	/soft	nfs	vers=4,hard,tcp,intr,timeo=20,retrans=10	0	0' >> $NEWROOT/etc/fstab

# wwsh object modify -a POSTSCRIPT=add_soft.sh <node>
```

renatopancheri:
i managed to use postscript and it worked fine, so i guess i will choose this path, thanks for the suggestion!

PR #225, opened 2020-03-03
Fixed wwmkchroot -g nogpgcheck option problems

NaohiroTamura:
This patch fixed the following three problems:

1. '"0$REPO_NOGPGCECK"' had two typos, 'REPO_' is unnecessary and
   'CECK' should be "CHECK", so it should be '"0$NOGPGCHECK"'.
2. '--setopt=gpgcheck=' with empty parameter caused error, so
   '$REPO_NOGPGCHECK' has to be either '0' or '1'
3. Inserted '--setopt=gpgcheck=$REPO_NOGPGCHECK' between '--add-repo'
   and '$i' caused '--add-repo' parameter error, so it should be
   '--setopt=gpgcheck=$REPO_NOGPGCHECK --add-repo $i'.

Signed-off-by: Naohiro Tamura <naohirot@jp.fujitsu.com>

### DIFF
--- a/vnfs/libexec/wwmkchroot/include-rhel
+++ b/vnfs/libexec/wwmkchroot/include-rhel
@@ -110,9 +110,9 @@ prechroot() {
     rm -f $CHROOTDIR/etc/yum.repos.d/*
     
     # Create new repos from comma-seperated list (using bash inline S&R)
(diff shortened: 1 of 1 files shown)


jmstover (approved):

jmstover:
LGTM. Thanks.

PR #226, opened 2020-03-04
Fix wwgetnodeconfig error when booting

fun2yang:
The wwgetnodeconfig and other scripts starts with wwget* will write logs into '/var/log/warewulf/provision/http.log'.  However, the 'provision' directory is not exist, which causes the error.  So, create it before exec the script.

### DIFF
--- a/provision/initramfs/init
+++ b/provision/initramfs/init
@@ -15,6 +15,7 @@ mkdir -p /dev
 mkdir -p /tmp
 mkdir -p /var
 mkdir -p /usr
+mkdir -p /var/log/warewulf/provision
 
 mount -t proc none /proc >/dev/null 2>&1
 mount -t sysfs none /sys >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


provision/initramfs/init
```diff
@@ -480,9 +480,8 @@ for i in /etc/warewulf/init/*; do
     fi
 done
 
-
+mkdir -p /var/log/warewulf/provision
```
jmstover: Can you move this up to just above the:
```
msg_blue "Now Booting Warewulf...\n"
```
line? Just make the directory at the start of the init file.

jmstover (changes requested):

provision/initramfs/init
fun2yang: OK, changed.

jmstover: Thanks.

jmstover (approved):

jmstover:
LGTM.

**Edit:**
Doh! Didn't mean to close it... *whistles innocently*

PR #229, opened 2020-03-06
Fixed efi_example.cmds swap partition problem

NaohiroTamura:
This patch fixed the following swap partition problem of efi.cmds
example for provisioning in UEFI mode.

 # cp /etc/warewulf/filesystem/examples/efi_example.cmds /etc/warewulf/filesystem/efi.cmds
 # wwsh -y provision set --filesystem=efi c1
 WARNING:  Command does not have at least 6 arguments, line: fstab 2 swap swap defaults 0

Signed-off-by: Naohiro Tamura <naohirot@jp.fujitsu.com>


jmstover (approved):

jmstover:
Thanks.

PR #232, opened 2020-03-11
Make NEWROOT early in init

jmstover:
Make sure the value of NEWROOT is created early in the /init process.


jmstover:
I somehow got in a state where `$NEWROOT` wasn't created, and it was trying to extract the VNFS from `/` in the initramfs ... That didn't work out so well.

bensallen (approved):

PR #236, opened 2020-04-05
Add support for dhcp shared networks

redsox38:
Our use case is that each rack in our cluster is a separate /26 using dhcp helper to forward requests to an off cluster ohpc management node. The purpose of the network cuts is to facilitate a smoother transition to ohpc 2.0 when it becomes available that doesn't require taking the entire cluster offline at once. The patch is based on this thread https://groups.google.com/a/lbl.gov/forum/#!topic/warewulf/Yx9OMqkyJKU and it adds a configuration directive "dhcp network = shared" to enable the behavior, otherwise it reverts to the default behavior.

### DIFF
--- a/provision/lib/Warewulf/Provision/Dhcp/Isc.pm
+++ b/provision/lib/Warewulf/Provision/Dhcp/Isc.pm
@@ -152,6 +152,7 @@ persist()
     my $ipaddr = $config->get("ip address") // $netobj->ipaddr($devname);
     my $netmask = $config->get("ip netmask") // $netobj->netmask($devname);
     my $network = $config->get("ip network") // $netobj->network($devname);
+    my $dhcp_net = $config->get("dhcp network") || "flat";
     my $config_template;
     my $dhcpd_contents;
     my %seen;
@@ -281,7 +282,7 @@ persist()
                     next;
                 }
 
-                if ($node_testnetwork ne $network) {
+                if (($dhcp_net eq "flat") and ($node_testnetwork ne $network)) {
                     &iprint("Skipping DHCP config for $nodename-$devname (on a different network)\n");
                     $dhcpd_contents .= "   # Skipping $nodename-$devname: Not on boot network ($node_testnetwork)\n";
                     next;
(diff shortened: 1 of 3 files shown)


renatopancheri:
Hi,

i read all the thread and i believe i have a very similar configuration with multiple clusters in different subnets, this patch might be very useful for me also.
My solution was to duplicate the master, put the copy in the second subnet and set the mysql address of the copy to the mysql of the original so that i can provision everything from the original master, i also created an additional event that creates a second dynamic_hosts containing only the nodes of the second subnet
I am interested to know if with this patch you've done  and/or with other stuff it's possible to achieve the same result but with reduced complexity.
If you need help with this or you have some tips to share please let me know!

redsox38:
This patch would include all hosts from all subnets in a single hosts file. If this patch is incorporated into the mainline, I was planning on expanding it by properly including all of the subnets from the hosts into the dhcp template rather than statically entering them as I am now. I'm not sure what the need to separate host files would be, as it seems to me that at that level of separation you might want a separate warewulf installation. However, I think a potentially more scalable way to handle it might be to optionally examine the subnet in HostsFile.pm and push out separate host files by subnet.

renatopancheri:
I see, i might try some stuff next days, i just noticed there is an unfinished/unused provision var called bootserver, that might be usefull

jmstover:
IIRC, `bootserver` was ment to be the DHCP `next server` . So... the DHCP request gets sent off to that other server. 

If you modify the checks that the node address is on the same network as the master when the DHCP/PXE files are written you can get this to work. Done it once before when didn't have any other way around it.

Warewulf was designed around the idea that there's a single provisioner per broadcast domain.  Normally, even if the database is shared, another master is setup if it's a different network. There's just stuff that's built around the idea that we're only going to be talking on, and controlling, a single network. Now... that may be a `10.0.0.0/8` network... but it's still a single network.

And sorry... haven't had a chance to look deep into this yet. :(  @redsox38 ... Can you rebase against the `development` branch though?

bensallen:
Closing in favor of #238.

PR #238, opened 2020-04-17
Add support for dhcp shared networks

redsox38:
Rebase #236  against upstream development branch

Our use case is that each rack in our cluster is a separate /26 using dhcp helper to forward requests to an off cluster ohpc management node. The purpose of the network cuts is to facilitate a smoother transition to ohpc 2.0 when it becomes available that doesn't require taking the entire cluster offline at once. The patch is based on this thread https://groups.google.com/a/lbl.gov/forum/#!topic/warewulf/Yx9OMqkyJKU and it adds a configuration directive "dhcp network = shared" to enable the behavior, otherwise it reverts to the default behavior.

### DIFF
--- a/provision/etc/provision.conf
+++ b/provision/etc/provision.conf
@@ -14,12 +14,14 @@ network device = eth1
 # select between 'isc' or 'dnsmasq'
 dhcp server = isc
 
-# Enable support for shared dhcp subnet blocks.
-# By default warewulf expects to share one flat network
-# If your SMS node is on a different subnet than
-# your compute nodes set this to shared.
-# Acceptable values are "shared" or "flat"
-dhcp network = flat
+# How should the DHCP subnet(s) be established?
+# direct - Only generate DHCP host reservations for nodes within the 
+# same subnet as the configured 'network device' or 
+# 'ip address' (default)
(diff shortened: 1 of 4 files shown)


jmstover:
Can you add an example `dhcp network` entry to `provision/etc/provision.conf`? Along with what you think the values should be?

If I'm reading it right, it looks like that should be set to `flat` if you want to keep current behavior. In which case, I would like that to be the default...

jmstover (dismissed):
Tentatively looks good. @bensallen ... I'll await your comments on it though. ;)

bensallen (changes requested):
Can we change the option values from `flat` and `shared` to perhaps `direct` and `relay`. Currently reading the config file I wouldn't actually know what this setting does.

We can describe the two options in the config file as:

Direct - Only generate DHCP host reservations for nodes within the same subnet as the configured 'network device' or 'ip address' (default)
Relay - Generate DHCP host reservations for all nodes in the Warewulf database without regard to subnet. Useful for DHCP relay configurations.

The code changes look fine otherwise.

bensallen (approved):
Looks good thanks!

PR #240, opened 2020-04-24
Add support for managing a pdsh genders file

redsox38:
This patch adds support for managing a pdsh genders file to Warewulf that reflects group membership in the genders file


jmstover (approved):
LGTM

PR #241, opened 2020-05-10
Leave disabled nodes out of genders file

redsox38:
Comment out and flag nodes marked as disabled from the genders file


jmstover (approved):

jmstover:
Thx.

PR #249, opened 2020-07-11
Replace /etc/os-release in chroot with RPM version

jmstover:
We create a minimal /etc/os-release file in the chroot to be able to build the chroot image. When an RPM package is installed, this file is seen, so the real /etc/os-release is written out as /etc/os-release.rpmnew ...

Make sure the RPM os-release is what we use.


bensallen (dismissed):
Seems reasonable.

bensallen (approved):
doh! good catch.

PR #250, opened 2020-07-29
Clarify meaning of "Signed-off-by" explanation.

alansill:
The "signed-off-by" block does not make it clear what has been signed off by the listed project developers. This addition adds a sentence to make this explanation clearer.

### DIFF
--- a/provision/3rd_party/GPL/README
+++ b/provision/3rd_party/GPL/README
@@ -10,7 +10,8 @@ must be respected and followed when redistributing Warewulf packages.
 
 The Warewulf developers on behalf of Lawrence Berkeley National Laboratory
 have contacted the upstream maintainers of these projects and notified them
-of our intentions.
+of our intentions. The Warewulf project will include binaries along with 
+the sources where necessary as explained above for the projects listed below.
 
 
 Signed-off-by:


tytso:
Please also remove the "Signed-off-by: " in Line 16 of the original file (Line 17 in 6fd0545).   Signed-off-by has a very specific legal meaning (see the Developer's Certification of Origin), and using specific legal terms without understanding what they mean is a Really Bad Idea --- especially when it implies that the people who were contacted had affirmatively agreed to something which is not clearly defined, and especially when other Warewulf project members started throwing around words like "plugins" without understanding what *that* term meant, to the point where I was contacted asking if I had given permission to do something very different from "Mere Aggregation", which does not need any kind of special permission from the developers involved.

bensallen:
Hi @tytso,

Perhaps some background context on why we do this would be useful. Lawrence Berkeley National Laboratory provided the project guidance when Warewulf was first being published. They were concerned with the project being BSD licensed and containing GPL licensed code that Warewulf may accidentally become a derivative work. They required we have a representative from each GPL licensed project review Warewulf's use and sign off.

While this project no longer has a direct connection to LBNL (eg. @gmkurtzer no longer works at LBNL), we still try to meet this requirement in good faith.

Note, I'm unsure what was discussed recently on the mailing list. Presumably, this was on OpenHPC's mailing list and not Warewulf's list?

Below is an example email for your reference that I've used for other sign offs, iPXE in this case. Hopefully this helps clarify what this sign off means.


> I'm emailing you as one of the current maintainers of iPXE. I'm part of a BSD licensed project named Warewulf (https://github.com/warewulf/warewulf3), which provides HPC systems provisioning. This project has been primarily developed and distributed by the Department of Energy and UC Regents.
> 
> We would like to include iPXE as part of the source and binary releases of Warewulf, replacing our use of Syslinux, to provide a more robust network boot loader and to add support for provisioning ARM64 nodes. iPXE will be maintained in a separate directory noted as 3rd_party/GPL in the project, and no code in that directory will be modified without either the patches also included in the 3rd_party/GPL directory and/or sent back upstream.
> 
> Note due to our license, it might be reasonable to assume that a vendor will want to add their non-distributable value additions to our software and distribute it themselves. They will need to be in compliance with all of the licenses that are part of Warewulf. As long as their modifications are only to the Warewulf/BSD code and not to the 3rd_party GPL packages, we don't believe they would be violating iPXE's GPLv2 license (as we interpret it).
> 
> If you *don't* feel as though this is reasonable, or not in compliance with your intentions of GPL'ing iPXE, then it means we will have to rely on the users to retrieve and build iPXE themselves. Either way is an acceptable path forward. However, we would like to make it as easy as possible for the users by including a version that we know has the features we need as well as provide a stack that has been tested and known to work.
> 
> If you *do* feel that it is reasonable for the project to distribute iPXE within Warewulf, Lawrence Berkeley National Laboratory's Tech Transfer department has asked us to contact the maintainers of all GPL 3rd_party software to request a statement that we can include in the distribution. This statement says that Warewulf is not a derivative work of yours and that you feel as though the redistribution of your software package as we have described is in alignment with your interpretation and intentions of the GPL. If you are okay with this, and agree to the statement here: https://github.com/warewulf/warewulf3/blob/master/provision/3rd_party/GPL/README, could you provide a quick reply that you agree to be listed under Signed-off-by? A pull request adding your sign-off would also work well.
> 
> Please let me know if you have any questions, comments, or need any additional detail.
> 
> For your reference the project's pull request that integrates iPXE can be found: https://github.com/warewulf/warewulf3/pull/46.
> 
> Thank you for your time and all your work on iPXE.


Note, a very similar email was sent to the parted maintainers. This email is based directly off an email that @gmkurtzer used to contact previous maintainers. Note, I can't speak to which projects preciously.

Regarding "Signed-of-by" itself. It's my opinion that since we don't mention the DCO, nor have a CLA in this project, that "Signed-of-by" is not specific to the DCO in our use.

Overall, I'm hesitant to make any change to this README outside of adding or removing projects without having all those named review the change.

Lastly, if you would like your name removed please feel free to open a PR or simply say so here. If you prefer this, the maintainers of the project will stop distributing E2fsprogs.

Thanks for all the work you do in the community.

Best,

Ben

alansill:
Thanks @bensallen for the clarifying comment. If you will inspect the proposed pull request that I put in, you can see that it is minimal and conveys the essence of what you have spelled out in your comment.

Can I suggest that the pull request be merged as compatible with the intent of the file here? All it does is spell out what the project does explicitly.

bensallen:
@alansill Note in the emails we have sent out is the line "agree to the statement here: https://github.com/warewulf/warewulf3/blob/master/provision/3rd_party/GPL/README". In my opinion, changing this README invalidates this agreement and thus in good faith we'd need to ask those that have signed off to agree to any modification.

I'm going to close this PR without merging. Hopefully this PR and its comments can serve as reference in the future.

PR #251, opened 2020-08-03
Add xfs support

sylviacy:
As the default fs on CentOS 7/8,
this patch introduces xfsprogs-5.0.0.tar.xz for help.


bensallen:
@sylviacy Lets remove the xfsprogs source and build from the PR and rely only on `--with-local-xfs`. 

`FIND_LOCAL(xfsprogs, mkfs.xfs)` needs to be added to https://github.com/warewulf/warewulf3/blob/development/provision/configure.ac#L160

sylviacy:
> @sylviacy Lets remove the xfsprogs source and build from the PR and rely only on `--with-local-xfs`.
> 
> `FIND_LOCAL(xfsprogs, mkfs.xfs)` needs to be added to https://github.com/warewulf/warewulf3/blob/development/provision/configure.ac#L160

Sure, I'll update this PR. Thanks.

PR #252, opened 2020-08-10
Fix depricated mkpath calls in Boostrap.pm

jlowellwofford:
Bootstrap.pm has a couple of `mkpath` calls in it.  `mkpath` has been deprecated, and should be replaced with `make_path`.  The `mkpath` calls fail on RHEL 8.1, Perl 5.26.3.  This fix has been tested on the same versions.


jmstover (approved):
LGTM. Thanks.

PR #253, opened 2020-08-10
Unescaped braces in Dnsmasq.pm

jlowellwofford:
On RHEL8.1 Perl 2.26.3 `Dnsmasq.pm` fails due to unescaped braces matching patterns like `%{IPADDR}` in the config template.  This PR escapes those braces.  Testing on the same versions.


jmstover (approved):
LGTM.

PR #254, opened 2020-08-10
Provision spec file dependency tweaks for RHEL 8.1

jlowellwofford:
This PR fixes two issues with the spec file for the provision service for RHEL 8.1:

1) the existing `device-mapper-devel` BuildRequires does not appear to be necessary for a successful build, and it requires pulling in the unsupported Code Ready Builder repository, so should probably be eliminated if unnecessary;
2) the build requires the `bsdtar` package which is not installed by default on RHEL 8.1, and is not specified in the spec file.


jlowellwofford:
Looks like this might need some conditionals to work on both 7 & 8.  I'll take a look.

jmstover:
@bensallen ... Was there a specific bsdtar requirement? Can't recall...

bensallen:
@jmstover we use bsdtar in the initramfs as it supports the pax stream format which supports xattrs. This lets images that are created with bsdcpio (eg. `cpio command = bsdcpio --quiet -o -H pax`) to include selinux labels, capabilities, and other xattrs. We're currently building libarchive (eg. bsdtar, et. al) in provision. If we want to include the system's local copy of bsdtar in the initramfs, then yes the spec would need to be updated with the aforementioned BuildRequire. Otherwise, I'm not sure why it would be needed as we have our own copy to build.

bensallen:
The device-mapper-devel requirement is for building parted with mapper support. It should probably stay in the spec.

jlowellwofford:
I can back out the device-mapper-devel change.  It's worth noting that it requires using an unsupported repo in RHEL8 to build.  But then again, getting mod_perl in RHEL8 requires much more questionable things than that...

bensallen (approved):
LGTM

PR #255, opened 2020-08-11
Clean up /dev mounting in init

### LINKED ISSUE #242: Duplicate /dev mounts
When a node is vnfs provisioned, it ends up with two entries for the `/dev` mount in `/proc/mounts` (etc).  After digging a bit, this happens because `provision/initramfs/capabilities/provision-vnfs/70-devtree` mounts dev in `$NEWROOT/dev`, but the one in the initramfs remains.

For almost everything, having a duplicate `/dev` mount doesn't hurt anything.  I ran into an issue specifically with running `libvirt` on a warewulf provisioned node, where the duplicate `/dev` entry confuses `libvirt` (specifically, because of a not-so-robust routine used in `libvirt`, it fails to mount `/dev/pts` in the domain namespace).

Here's one way this could be managed:

1. I suspect `provision/initramfs/capabilities/provision-vnfs/70-devtree` may no longer be needed.  It has the ability to populate a `/dev` for linux that doesn't support `devtmpfs`, but `devtmpfs` has been in the kernel for the last decade.  Maybe just get rid of it.
2. Since `initramfs/init` handles mounting `/proc`, `/sys`, and `/dev` initially, it makes sense to do the following:
```sh
mount --move /proc "$NEWROOT/proc"
mount --move /sys "$NEWROOT/sys"
mount --move /dev "$NEWROOT/dev"

exec switch_root "$NEWROOT" /sbin/init
```

I've tested this, and it resolved this issue for a centos7 vnfs image.

Happy to create a PR if this seems reasonable.

jlowellwofford:
Resolves #242 

This PR cleans up how `/dev` is handled in init.  It does the following:

1) Make sure `devtmpfs` is mounted before anything touches `/dev` (e.g. `> /dev/null`).
2) `mount --move` `/dev`, `/sys`, and `/proc` just before calling `switch_root`.
3) Remove deprecated `70-devtree`, since all modern kernels should support `devtmpfs`.

Together, these fixes also resolve the issue noted in #242 which results in duplicate `/dev` mounts.


bensallen (approved):
LGTM

PR #256, opened 2020-08-19
Add xfs support

sylviacy:
As the default fs on CentOS 7/8,
this patch introduces --with-local-xfsprogs for local building.


sylviacy:
@bensallen 
Hi, any comments?

bensallen (approved):
Looks good. Thanks!

PR #269, opened 2020-11-23
Remove /dev/log after syslog is shutdown

bensallen:
On SUSE15 at least, /dev/log is a symlink to /run/systemd/journal/dev-log. With WW's initrd /dev/log socket left over, this symlink is never created and /dev/log isn't usable, eg:

```
logger: socket /dev/log: Connection refused
```


bensallen:
Confirmed on RHEL7 that /dev/log is created after switch_root if its removed.

PR #271, opened 2021-01-12
Add use of gzip command in wwbootstrap as defined in vnfs.conf

bensallen:
Add use of gzip command in wwbootstrap as defined in vnfs.conf. This allows the use of pigz in wwbootstrap, which can drastically speed up processing of large bootstraps.


jmstover (dismissed):
looks good once the CI passes

PR #290, opened 2021-10-14
SUSE netconfig

e4t:
Number of patches to properly configure the network with warewulf for 'wicked' on SUSE systems.


bensallen:
Thanks for these. Fixing up Suse ifcfg files has long been on my todo list.

bensallen (approved):

PR #291, opened 2021-10-14
Add /lib/modules/$opt_kversion/sysctl.conf to initfs if present

e4t:
Kernels may ship their own sysctl.conf with settings specific
to the respective kernel version. This is shipped with the
kernel itself in the module directory.


bensallen (dismissed):

bensallen:
@e4t can you point me to docs or examples where kernels are getting deployed with their own sysctl.conf? Curious what its being used for and a quick search doesn't find anything.

e4t:
> @e4t can you point me to docs or examples where kernels are getting deployed with their own sysctl.conf? Curious what its being used for and a quick search doesn't find anything.

Well, I had wondered about this as well when I stumble over it. So I've asked Takashi Iwai - a senior kernel developer at SUSE - here is the conversation:

> _Got a kernel-related question:
> It looks like Tumbleweed kernel now ships a file
> /usr/lib/modules/`uname -r`/sysctl.conf_

_The sysctl config per kernel is not new, it's used to be in
/boot/sysctl*, while it's now moved to /usr/lib/modules/* due to the
recent UsrMerge action.  There is still a symlink in /boot for
compatibility, too._

> _Can you tell me what the purpose of this file is?
> Does it contain sysctl settings specific to this
> kernel - which therefore cannot be provided in
> /etc/sysctl.d/?
> Is this SUSE specific or a general new feature?_

_It's not really specific to SUSE but a generic feature._

_That contains the extra sysctl configuration for each kernel flavor.
We don't use /etc/sysctl.d just because we want to install multiple
versions and the contents may change that may conflict in each
different version._

_Takashi_

provision/initramfs/capabilities/provision-vnfs/70-kernelmodules
```diff
@@ -33,4 +33,8 @@ if [ -x "${NEWROOT}${DEPMOD}" ]; then
     chroot "${NEWROOT}" /sbin/depmod -a
 fi
 
+if [ -e "${NEWROOT}"/lib/modules/"${KVERSION}"/sysctl.conf ]; then
+    ln -s /lib/modules/"${KVERSION}"/sysctl.conf" "${NEWROOT}"/boot/sysctl.conf-"${KVERSION}"
```
bensallen, line 37: Thinking we should check if `"${NEWROOT}"/boot/sysctl.conf-"${KVERSION}"` exists before creating this symlink.

bensallen:
@e4t thanks for the background.

provision/initramfs/capabilities/provision-vnfs/70-kernelmodules
e4t: Indeed. Thanks for catching!
Let me see if I can push an updated patch. I will fix the changelog entry to better explain the rationale.

bensallen (approved):

PR #292, opened 2021-10-14
initramfs: Going forward, /bin or /sbin are mere links to /usr - cate…

e4t:
…r for this

Create relative links to help installation remove before creating the image.
This makes things easier once this concept is used by busybox et.al. -
which is the case already when using distro packages.


provision/initramfs/Makefile.am
```diff
@@ -145,6 +145,8 @@ rootfs: busybox e2fsprogs xfsprogs libarchive parted curl
 	mkdir rootfs/usr
 	mkdir rootfs/lib
 	mkdir rootfs/lib64
+	ln -s ../bin rootfs/usr/bin
```
jmstover: Maybe I'm missing something here... but isn't this creating a symlink *at*  rootfs/usr/bin pointing to ../bin ? 

And if /bin, /sbin are the links ... why would you make rootfs/usr/bin the link?

provision/initramfs/Makefile.am
```diff
@@ -203,6 +205,8 @@ rootfs: busybox e2fsprogs xfsprogs libarchive parted curl
 	ln -s transporthandler rootfs/sbin/wwgetscript
 	ln -s transporthandler rootfs/sbin/wwgetnodeconfig
 	ln -s transporthandler rootfs/sbin/wwgetfiles
+	rm -f rootfs/usr/bin
```
jmstover: If these are the actual location , why would you try removing them?

jmstover (commented):
I'm morally against the symlink of /bin -> /usr/bin and /sbin -> /usr/sbin ... I *still* do NFS mounted /usr directories when I can.

provision/initramfs/Makefile.am
e4t: Yes, it's the other way around. It links /usr/[s]bin to /bin. The direction is pretty much irrelevant as it is just for the purpose of the initramfs. This is short lived and will not be used once the 'main' system is up and running.
Why did I do it this way? Maybe I'm disliking  this 'usr merge' as much as you are ...
I'm with you regarding NFS-mounting /usr. Luckily this is not an issue here. The reasons I've heard  behind this 'merge' did not sound very compelling, either. My original changelog comment reflected this feeling, I've changed it to a more neutral wording.

provision/initramfs/Makefile.am
e4t: Right. Something got lost from my original patch in rebasing. Let me push it again ...

e4t: I've pushed the correct version now. The links need to be recreated to point to the absolute location of /bin and /sbin, not the relative one under rootfs/.

bensallen:
@e4t Is ensuring the various busybox links are in /sbin and /bin just for convenience, or is this fixing some specific issue?

e4t:
> @e4t Is ensuring the various busybox links are in /sbin and /bin just for convenience, or is this fixing some specific issue?

It is a little more than just convenience. There is a trend to deprecate /bin and /sbin and move everything to /usr/[s]bin. Presently, distros create compatibility symlinks from /[s]bin to /usr/[s]bin to not break scripts that have this hard-coded. Most upstream packages have not yet followed this trend so distros do some scripting to implement this after running 'make install'. 
Currently, busybox doesn't have this implemented either, however, at SUSE we do not used the warewulf provided version of busybox but our busybox package to build the initramfs for our warewulf package. On openSUSE:Tumbleweed, this package has been converted already. To work under all circumstances, regardless of which version of busybox is used and how this has been built, I've made this change.

bensallen (approved):

PR #312, opened 2022-07-21
Drop the hwprefix node attribute and the use of DHCP client identifer for IB

bensallen:
It seems with UEFI based PXE booting over IB, both UEFI and iPXE no longer attempt to use the DHCP client identifier. While its technically more correct to use the GUID of an IB adapter, it seems its no longer practical to do so.

1. Dropping the use of hwprefix
2. Updating init in the initramfs to calculate the 48-bit HWADDR from the the 20 byte address instead of the 64-bit GUID.

From

```
HWADDR=`expr substr $HWADDR 37 23`
```

To

```
HWADDR=`expr substr $HWADDR 37 8`:`expr substr $HWADDR 52 8`
```


jmstover (approved):
Looks good from my side.

*Edit:* Can't approve anymore though.

bensallen:
@jmstover Thanks. Re-invited you to the organization.

PR #319, opened 2023-03-24
provision: Update ipxe to latest master (09e8a15)

martin-g:
While testing Warewulf 3.10.0 provisioning in OpenHPC we experienced issues with iPXE 2265a65 - the compute nodes didn't boot.

According to https://ipxe.org/faq:
```
Which version of iPXE should I use?

iPXE uses a rolling release model, in which every commit is intended to be production-ready. You should always use the latest code. 
```

The removed patch is already applied with https://github.com/ipxe/ipxe/commit/c6901792f009cfd824707724b687e99edd4c8ecd


bensallen:
Thanks LGTM

PR #324, opened 2024-03-27
Make Build reproducible

e4t:
This pull request supersedes pull request #323:
There is a project under way to convince and help software projects to make their builds reproducible:
 https://reproducible-builds.org/
openSUSE Factory is participating in this by attempting to make the entire build reproducible: https://en.opensuse.org/openSUSE:Reproducible_Builds. If packages contain files that may change between builds, these packages require changes. Since warewulf3 includes CPIO archives of an initramdisk there is a bit more effort involved in making the build reproducible: not only the file content but also the time stamps of the files in these archives matter. Moreover, cpio needs to be run with the option `--reproducible`.
Out tests have shown that by setting the access and modification time to a fixed reference value using touch is sufficient to make the resulting CPIO archives reproducible.


e4t:
The build errors in the CI seem to be unrelated to the patch.

